PR #155, opened 2014-09-18
Allow custom CSS formatting in published HTML

jacebrowning:
Migrated from #136

-----

- [ ] add custom CSS as an example


jacebrowning:
Someone else could add this feature in the future.

PR #192, opened 2015-01-29
Add an option to skip documents during validation

### LINKED ISSUE #191: Add option to suppress messages for selected documents
We're developing multiple documents/levels and are transitioning certain ones under formal change control.

It would be helpful if we could run the `doorstop` validation, but suppress the `ERROR`|`WARNING`|`INFO` for select documents.

Example:

```
Feature List [FL]
  |
  -- System Requirements Specification [SYSRS]
    |
    ---- Software High-Level Requirements [SWHLR]
    |
    ---- Hardware High-Level Requirements [HWHLR]
```

The `FL` and `SYSRS` are reviewed and approved by the customer. Work is beginning on `SWHLR` and `HWHLR`. We need to track/fix `FL`|`SYSRS`, but `SWHLR`|`HWHLR` are a little more free-flowing.

I'd like to run `$ doorstop` from my `Makefile` and CI and only generate errors when `FL` and/or `SYSRS` are affected. Also since `FL`|`SYSRS` are held to a higher level of integrity, I'm also elevating `WARNING` -> `ERROR`.

An example command line call could be:
`$ doorstop --no-level-check --error-all --suppress SWHLR --suppress HWHLR`

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8113433-add-option-to-suppress-messages-for-selected-documents?utm_campaign=plugin&utm_content=tracker%2F537563&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F537563&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jacebrowning:
Fixes #191

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Changelog
 
 - Fixed a bug checking levels across inactive items.
 - Added error message for all IO errors.
+- Added '--skip' options to disable documents during validation.
 
 0.8.3 (2014/10/10)
 ------------------
--- a/doorstop/__init__.py
+++ b/doorstop/__init__.py
@@ -1,5 +1,7 @@
 """Package for doorstop."""
 
+import sys
+
 __project__ = 'Doorstop'
 __version__ = '0.9-dev'
 
@@ -11,7 +13,6 @@ DESCRIPTION = "Requirements management using version control."
 
 MIN_PYTHON_VERSION = 3, 3
 
-import sys
 if not sys.version_info >= MIN_PYTHON_VERSION:  # pragma: no cover (manual test)
     exit("Python {}.{}+ is required.".format(*MIN_PYTHON_VERSION))
 
--- a/doorstop/cli/commands.py
+++ b/doorstop/cli/commands.py
@@ -38,7 +38,7 @@ def run(args, cwd, error, catch=True):  # pylint: disable=W0613
 
         # validate it
         utilities.show("validating items...", flush=True)
-        valid = tree.validate()
+        valid = tree.validate(skip=args.skip)
 
     if not success:
         return False
--- a/doorstop/cli/main.py
+++ b/doorstop/cli/main.py
@@ -56,6 +56,8 @@ def main(args=None):  # pylint: disable=R0915
                         help="do not check for suspect links")
     parser.add_argument('-W', '--no-review-check', action='store_true',
                         help="do not check item review status")
+    parser.add_argument('-s', '--skip', metavar='PREFIX', action='append',
+                        help="skip a document during validation")
     parser.add_argument('-w', '--warn-all', action='store_true',
                         help="display all info-level issues as warnings")
     parser.add_argument('-e', '--error-all', action='store_true',
--- a/doorstop/core/base.py
+++ b/doorstop/core/base.py
@@ -111,9 +111,10 @@ class BaseValidatable(object, metaclass=abc.ABCMeta):  # pylint:disable=R0921
 
     """Abstract Base Class for objects that can be validated."""
 
-    def validate(self, document_hook=None, item_hook=None):
+    def validate(self, skip=None, document_hook=None, item_hook=None):
         """Check the object for validity.
 
+        :param skip: list of document prefixes to skip
         :param document_hook: function to call for custom document
             validation
         :param item_hook: function to call for custom item validation
@@ -123,7 +124,7 @@ class BaseValidatable(object, metaclass=abc.ABCMeta):  # pylint:disable=R0921
         """
         valid = True
         # Display all issues
-        for issue in self.get_issues(document_hook=document_hook,
+        for issue in self.get_issues(skip=skip, document_hook=document_hook,
                                      item_hook=item_hook):
             if isinstance(issue, DoorstopInfo) and not settings.WARN_ALL:
                 log.info(issue)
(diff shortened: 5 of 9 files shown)


robertberrington:
Looks like it's working pretty good. There are few residual messages I wonder if they should also be suppressed:

**Example 1:**
SYSRS-0680 has children, but only MEHLR is under change control, but should NOT link to SYSRS-0680 (a Software slanted requirement)
```
$ doorstop --skip CLHLR --skip HWHLR --skip SWHLR
...
WARNING: SYSRS: SYSRS-0680: no links from child document: CLHLR
WARNING: SYSRS: SYSRS-0680: no links from child document: HWHLR
WARNING: SYSRS: SYSRS-0680: no links from child document: MEHLR
WARNING: SYSRS: SYSRS-0680: no links from child document: SWHLR
...
```
Adding `--no-child-check` suppresses ALL missing parent -> child links. Is there a middle ground where we can suppress this for "skipped" documents?

Also, without making a "dummy" MEHLR item to point blindly at intentionally un-developed SYSRS, how do I flag these relationship in doorstop so I can elevate `WARNING` -> `ERROR` as part of my CI once I'm under change control?

**Example 2:**
SWHLR-0093 is `active = FALSE` and points to a SYSRS item. Perhaps this should be suppressed for documents skipped?
```
$ doorstop --skip CLHLR --skip HWHLR --skip SWHLR
...
WARNING: no item with UID: SWHLR-0093
...
```

So ... other than those two observations ... looks great!

Perhaps consider making the 1st example above a separate issue ... I think that one may be a large "unaddressed" issue we haven't wrestled with enough.

robertberrington:
@jacebrowning ... I was talking with Javis concerning the issue regarding "Parent-Child" links (or absence).

What about adding a property like `item.allocation`. This could be a space-delimited space to list the `child` document `ID`'s that should be linked. This is a "design-step" following the development of the "requirement", but would be a way to validate that "children" requirements exist where they are expected.

In a current project we are setting up a design-review meeting to discuss each requirements allocation to HW | SW | CL | ME. This is also meant to make sure we've addressed all the requirements in the design. We could just as easily populate this field directly in `doorstop` with a list of ID's where lower level requirements are expected.

jacebrowning:
@robertberrington This commit should fix your **Example 1**. I agree these "reverse lookup" warnings should be suppressed if the document is skipped.

I think that **Example 2** should always be a warning even if a document is skipped. This warning is telling you that you're linking to an item that essentially doesn't exist.

Regarding your second comment, if this is still desirable after the above commit, please submit a separate issue to discuss further.

robertberrington:
@jacebrowning I tested the commit: `1fab5bbbb14b71422450955eec8282a5b9fddf50`

I'm still getting the "reverse look-up" warnings for skipped documents.

```
$ doorstop --no-level-check --skip HWHLR --skip SWHLR --skip CLHLR
...
WARNING: SYSRS: SYSRS-0680: no links from child document: CLHLR
WARNING: SYSRS: SYSRS-0680: no links from child document: HWHLR
WARNING: SYSRS: SYSRS-0680: no links from child document: MEHLR
WARNING: SYSRS: SYSRS-0680: no links from child document: SWHLR
...
```

I think since `MEHLR` is not skipped, the `WARNING: SYSRS: SYSRS-0680: no links from child document: MEHLR` is the only one that should remain.

Also, just to double-check your language/thinking:
>
I think that Example 2 should always be a warning even if a document is skipped. This warning is telling you that you're linking to an item that essentially doesn't exist.
>
Just to clarify, I'm getting a warning on the child that is `active=FALSE` ... if the child is essentially turned OFF, should the warning be suppressed? I could see a nuance to the validation where you will throw a `WARNING: ... no links from child document ...` if there are no `active=TRUE` items linked.

The way you worded your justification, I agree ... if the parent is `active=FALSE`, you should throw a `WARNING` for the child that it is linking to an turned-OFF/inactive/deleted requirement.

I'm wondering if the `no links from child document` warning is suppressed if the a Parent <- Child link exists, but the Child is `active=FALSE`?

By the way, I really like the `active=FALSE` feature. It allows me to turn requirements ON/OFF as part of feature planning so that I can maintain a main-line of development and manage `active` with a flag instead of a [repo]-delete.

jacebrowning:
@robertberrington Let me know if that looks good for your **Example 1**.

I think you'll find that in some of your **Example 2** cases, it's not actually causing your builds to fail, but just printing a warning. Either way, I think that should be a separate discussion.

robertberrington:
The works well. The only issue I have is when I elevate `WARNING` to `ERROR` using the `--error-all` flag, the `WARNING: SYSRS: SYSRS-0615: no links from child document: MEHLR` causes the target to fail.

Since there are other documents in development, they will eventually cover these SYSRS items.

Would it make sense that the `ERROR: SYSRS: SYSRS-0615: no links from child document: MEHLR` would remain a `WARNING` when the `--error-all` is used ... like `WARNING: no item with UID: SWHLR-0057` (where SWHLR is `active=FALSE`, but it links to an active SYSRS)?

Frankly, I like that a `WARNING` -> `ERROR` is thrown in this case, except I've looked at all these instances and know they are intentional.

The feature requested in #194 would potentially address this.

PR #197, opened 2015-07-01
Added object and test suite for Mercurial plug-in

tjasz:
Though not as widespread as Git, as a distributed version control system, Mercurial is becoming quite popular. A relatively similar command-line interface to Git makes adding a plug-in to Doorstop quite simple, but eases use for developers who prefer Mercurial.


jacebrowning:
Thanks for the addition!

PR #206, opened 2016-06-17
Report the item id that generated the exception

michaelnt:
If the yaml file of an item is edited directly then it can cause an exception when later doorstop commands are run, e.g. publish. This change displays the UID of the item that caused the exception making it easier to fix.

### DIFF
--- a/doorstop/core/document.py
+++ b/doorstop/core/document.py
@@ -193,7 +193,11 @@ class Document(BaseValidatable, BaseFileObject):  # pylint: disable=R0902
                 else:
                     self._items.append(item)
                     if reload:
-                        item.load(reload=reload)
+                        try:
+                            item.load(reload=reload)
+                        except Exception as e:
+                            print("!! Exception loading %s" % item)
+                            raise e
                     if settings.CACHE_ITEMS and self.tree:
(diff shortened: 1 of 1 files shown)


doorstop/core/document.py
```diff
@@ -193,7 +193,11 @@ def _iter(self, reload=False):
                 else:
                     self._items.append(item)
                     if reload:
-                        item.load(reload=reload)
+                        try:
+                            item.load(reload=reload)
+                        except Exception as e:
+                            print("!! Exception loading %s" % item)
```
jacebrowning: I think this might be better as: `log.warning("Unable to load: %s", item)`

doorstop/core/document.py
```diff
@@ -193,7 +193,11 @@ def _iter(self, reload=False):
...
+                            raise e
```
jacebrowning: I believe this can simple be a bare `raise` statement and you can remove `as e` from above.

michaelnt:
Ok, should the print be a log.err?

jacebrowning:
That would work too.

michaelnt:
I think an error probably makes more sense as the program stops running at this point. Elsewhere WARNING is used for missing links, text etc. but the program continues to run.

PR #207, opened 2016-06-17
Support repositories where the property svn:ignore isn't set

michaelnt:
If svn:ignore isn't set then this function should return an empty list rather than None


jacebrowning:
Thanks for fixing that!

PR #208, opened 2016-06-17
Support the use of markdwn tables in the text descriptions

michaelnt:
I've made these changes so that I can use markdown tables in text field. I don't think the code currently support any multi line features. I expect I should add some tests before this could get merged.

Tables can be formatted like this.

| Headin1 | Heading2 |
| --- | --- |
| Row 1 | some number |

### DIFF
--- a/doorstop/core/publisher.py
+++ b/doorstop/core/publisher.py
@@ -294,7 +294,7 @@ def _lines_markdown(obj, linkify=False):
             # Text
             if item.text:
                 yield ""  # break before text
-                yield from item.text.splitlines()
+                yield item.text
 
             # Reference
             if item.ref:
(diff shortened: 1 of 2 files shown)


doorstop/core/publisher.py
```diff
@@ -428,7 +428,7 @@ def _lines_html(obj, linkify=False, charset='UTF-8'):
         yield '</head>'
         yield '<body>'
     text = '\n'.join(_lines_markdown(obj, linkify=linkify))
-    html = markdown.markdown(text, extensions=['extra', 'nl2br', 'sane_lists'])
+    html = markdown.markdown(text, extensions=['extra', 'sane_lists', 'tables'])
```
jacebrowning: Are the `nl2br` and `tables` extensions incompatible with each other?

michaelnt:
No but the line breaks in the text fields are no longer striped so become extra <br/> tags which looks odd.

jacebrowning:
I believe this has been superseded by https://github.com/jacebrowning/doorstop/pull/214.

Issue #141, opened 2014-08-22
Add option to warn if all children don't link to every parent


robertberrington:
In "light" requirements, children only link to parents where necessary. However, in a formal requirements setting, you might consider forcing all children to link to parents, even if it is a placeholder justifying why the parent does NOT impact the child.

michaelnt:
For instance in the requirements for doorstop the children of REQ are TUT and LLT, every REQ must be linked from either TUT OR LLT but there is no warning for items that aren't linked from BOTH.

WARNING: REQ: REQ001: no links from document: TUT
WARNING: REQ: REQ016: no links from document: LLT
WARNING: REQ: REQ017: no links from document: LLT
WARNING: REQ: REQ008: no links from document: TUT
WARNING: REQ: REQ009: no links from document: TUT
WARNING: REQ: REQ014: no links from document: TUT
WARNING: REQ: REQ015: no links from document: TUT

jacebrowning:
This feature is now available for testing using the --strict-child-check` option with `Doorstop==1.1b1`.

PR #210, opened 2016-12-09
Warn if an item doesn't have a link from every child document

michaelnt:
Fixes issue #141.

Let me know what you'd like changed to get this merged.


doorstop/cli/main.py
```diff
@@ -52,6 +52,8 @@ def main(args=None):  # pylint: disable=R0915
                         help="do not validate external file references")
     parser.add_argument('-C', '--no-child-check', action='store_true',
                         help="do not validate child (reverse) links")
+    parser.add_argument('-Z', '--strict-child-check', action='store_true',
+                        help="Strictly validate (reverse) links from every child document")
```
jacebrowning: To me, a help message like "require child (reverse) links from every document" is more clear.

doorstop/settings.py
```diff
@@ -30,6 +30,7 @@
 CHECK_LEVELS = True  # validate document levels during validation
 CHECK_REF = True  # validate external file references
 CHECK_CHILD_LINKS = True  # validate reverse links
+CHECK_CHILD_LINKS_STRICT = False  # Strictly validate reverse links from all the child documents
```
jacebrowning: To me, a comment like "require reverse links from every child document" is more clear.

doorstop/core/item.py
```diff
@@ -652,7 +653,7 @@ def _get_issues_tree(self, tree):
         if settings.REFORMAT:
             self._data['links'] = identifiers
 
-    def _get_issues_both(self, document, tree, skip):
+    def _get_issues_both(self, document, tree, skip, children):
```
jacebrowning: Is there any way this could be accomplished without passing `children` as a argument? This method already has a reference to the tree, so the child documents could be determined inside the `if settings.CHECK_CHILD_LINKS_STRICT:` block. I'm concerned about this breaking other functionality and showing down validation in cases were `CHECK_CHILD_LINKS_STRICT` is not needed.

michaelnt:
I changed the help string and settings comment. 

I removed the children argument but unfortunately document.tree points to the root of the tree so I added document.children that calls self.tree.get_prefix_of_children() to return the tree below a document. Perhaps that code should be in just tree or document.

doorstop/core/item.py
```diff
@@ -664,7 +665,8 @@ def _get_issues_both(self, document, tree, skip):
         if settings.CHECK_CHILD_LINKS and self.normative:
             items, documents = self._find_child_objects(document=document,
                                                         tree=tree,
-                                                        find_all=False)
+                                                        find_all=True)
```
michaelnt: Should find_all only be run if CHECK_CHILD_LINKS_STRICT is true to speed things up?

jacebrowning: Yeah, I think so.

doorstop/core/tree.py
```diff
@@ -437,7 +438,12 @@ def get_issues(self, skip=None, document_hook=None, item_hook=None):
         if not documents:
             yield DoorstopWarning("no documents")
         # Check each document
-        for document in documents:
+        trees = [self]
```
jacebrowning: What is this logic doing? I'm a bit confused to see a variable named `trees` -- is there more than one?

doorstop/core/item.py
michaelnt: I've made that change

doorstop/core/tree.py
michaelnt: Sorry that isn't needed anymore and I've removed this change

michaelnt:
Rather than updating document.children when ever it is read it is now set when the tree is modified. I used the method _place, is that the right place to put the code?

doorstop/core/tree.py
```diff
@@ -477,6 +479,16 @@ def by_uid(row):
...
 
+    def get_prefix_of_children(self, document):
+        """Return the prefixes of the children of this document."""
+        for child in self.children:
+            if child.document == document:
+                children = [c.document.prefix for c in child.children]
+                document.children = children
+                return
+        children = [c.document.prefix for c in self.children]
+        document.children = children
```
michaelnt: OK, thats changed

jacebrowning:
I'm going to make a beta release for this. Feel free to make a new PR with any tweaks.

Issue #211, opened 2016-12-19
publish option no-body-levels should also drop levels from headings


michaelnt:
I have been publishing requirements to markdown and then using pandoc to convert to word. In word I have numbered headings so I want to remove the levels from the published markdown. The option no-body-levels leaves the level markers on the headings.

Should there be a new option that also drops the levels from the headings, or should the existing option be modified?

Perhaps --no-levels=[all|body]

jacebrowning:
I like `--no-levels=[all|body]`.

jacebrowning:
This is available in `Doorstop==1.2b1`.

PR #212, opened 2016-12-19
Add option to omit levels from all items or only non heading items

michaelnt:
PR to close issue #211.

I left in the -L option but removed the --no-body-levels option so shouldn't break existing use too much.

### DIFF
--- a/doorstop/cli/test/test_all.py
+++ b/doorstop/cli/test/test_all.py
@@ -699,11 +699,16 @@ class TestPublish(TempTestCase):
 
     def test_publish_document_without_body_levels(self):
         """Verify 'doorstop publish' can create output without body levels."""
+        self.assertIs(None, main(['publish', 'tut', '--no-body-levels']))
+        self.assertFalse(settings.PUBLISH_BODY_LEVELS)
+
+    def test_publish_document_no_body_levels(self):
+        """Verify 'doorstop publish' can create output without body levels."""
         self.assertIs(None, main(['publish', 'tut', '--no-levels=body']))
         self.assertFalse(settings.PUBLISH_BODY_LEVELS)
 
-    def test_publish_document_without_heading_levels(self):
-        """Verify 'doorstop publish' can create output without heading levels."""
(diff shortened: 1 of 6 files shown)


jacebrowning:
Looks pretty good so far. Wamt to add a test here? https://github.com/jacebrowning/doorstop/blob/46b108d912673a896e5d830d78b6681422137b77/doorstop/core/test/test_publisher.py#L14

doorstop/cli/test/test_all.py
```diff
@@ -699,9 +699,15 @@ def test_publish_document_without_child_links(self):
 
     def test_publish_document_without_body_levels(self):
         """Verify 'doorstop publish' can create output without body levels."""
-        self.assertIs(None, main(['publish', 'tut', '--no-body-levels']))
```
jacebrowning: To ensure that no one's existing workflow breaks, `--no-body-levels` should still be support and this should become two tests, one for `--no-levels=body` and another for the legacy method.

doorstop/cli/utilities.py
```diff
@@ -112,8 +112,11 @@ def configure_settings(args):
     # Parse `publish` settings
     if hasattr(args, 'no_child_links') and args.no_child_links is not None:
         settings.PUBLISH_CHILD_LINKS = args.no_child_links is False
-    if hasattr(args, 'no_body_levels') and args.no_body_levels is not None:
-        settings.PUBLISH_BODY_LEVELS = args.no_body_levels is False
+    if hasattr(args, 'no_body_levels'):
+        settings.PUBLISH_BODY_LEVELS = not args.no_body_levels
```
jacebrowning: Please add a deprecation warning here so this option can be safely removed in the 2.0 release:

```python
msg = "'--no-body-levels' option will be removed in a future release"
warnings.warn(msg, DeprecationWarning)
```

michaelnt: I've made that change, the warning doesn't get printed by default. 

I could see the warning and some other warnings if I run python -Wd

PR #214, opened 2017-01-17
Text is no longer re-wrapped at sentence boundaries

### LINKED ISSUE #193: Fenced markdown is reformatted
This text should not be reformatted:

``````
```
# A heading
```
``````
##

michaelnt:
Text is no longer re-wrapped at sentence boundaries and new lines and the beginning and end are removed. Should close #193. 

If a line of text starts with spaces then the literal style yaml looks like this, which I thought was a bug, but it is indicating the number of spaces used to indent the text.
|2
   hello world


jacebrowning:
Please test this in `Doorstop==1.2b1`.

PR #215, opened 2017-01-18
 Remove trailing white space from every line of text

michaelnt:
pyyaml stops saving text as literal if there is a trailing space before a new line. The code previously removed trailing white space from the last line, but now it removes trailing white space from every line.


jacebrowning:
This change is available in `Doorstop==1.2b2`

PR #218, opened 2017-01-25
The markdown module no longer converts newlines to breaks

### LINKED ISSUE #217: Don't convert newlines to <\br> when publishing html
Lines of text in the yaml file may have been manually wrapped at some column width that doesn't match the column width of the published html, resulting in ugly extra line breaks. Previously I think these lines breaks would have been removed by the sentence boundary re-wrapping code. This behavior doesn't conform to normal markdown and can be worked around by publishing to markdown and then using pandoc to convert to html.

Can we turn off the nl2br option when publishing to html?

michaelnt:
Removed the extension nl2br so that new lines are no longer converted to breaks, closes #217.

### DIFF
--- a/doorstop/core/publisher.py
+++ b/doorstop/core/publisher.py
@@ -95,7 +95,7 @@ def _index(directory, index=INDEX, extensions=('.html',), tree=None):
     if filenames:
         path = os.path.join(directory, index)
         log.info("creating an {}...".format(index))
-        lines = _lines_index(filenames, tree=tree)
+        lines = _lines_index(sorted(filenames), tree=tree)
         common.write_lines(lines, path)
     else:
         log.warning("no files for {}".format(index))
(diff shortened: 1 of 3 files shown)


doorstop/core/test/files/index.html
```diff
@@ -130,8 +130,8 @@
 <p>
 <ul>
 <li> <a href="index2.html">index2</a> </li>
-<li> <a href="published.html">published</a> </li>
 <li> <a href="published2.html">published2</a> </li>
+<li> <a href="published.html">published</a> </li>
```
jacebrowning: I've noticed that this file keeps changing on your PRs, so I assume the issue is OS-specific.

Do you think you could add a fix that ensures these filenames are sorted according to Python sort order?

michaelnt: I added the sort and those files are no longer changed

PR #220, opened 2017-02-02
Parent links are now links in markdown documents

michaelnt:
When publish all of the documents to markdown I think the parent links should be linkified, just as if we are publishing html.

If the markdown is later converted to html with pandoc then the links word correctly


michaelnt:
Hmm the attrs are now being added to the headings, e.g.

## 1.1 HL202 {: HL202 }

PR #221, opened 2017-02-02
Attributes are now added to the first line of a heading item

michaelnt:
Even if heading has multiple lines the attribute is now correctly added to the end of the first line, which becomes the heading text. The remaining lines can be used for some descriptive introduction under the heading.

### DIFF
--- a/doorstop/core/publisher.py
+++ b/doorstop/core/publisher.py
@@ -282,15 +282,15 @@ def _lines_markdown(obj, linkify=False):
         level = _format_level(item.level)
 
         if item.heading:
-
+            text_lines = item.text.splitlines()
             # Level and Text
             if settings.PUBLISH_HEADING_LEVELS:
-                standard = "{h} {l} {t}".format(h=heading, l=level, t=item.text)
+                standard = "{h} {l} {t}".format(h=heading, l=level, t=text_lines[0])
             else:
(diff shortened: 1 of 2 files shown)


jacebrowning:
What is the purpose of having a heading with multiple lines in it? In my opinion, non-normative section descriptions should be their own items. This mirrors the behavior of other requirements management tools.

michaelnt:
I'm using pandoc, jquery and bootstrap to get a html output with a table of contents in a navbar. When you look at the table of contents you quickly want real names for all of the H1 and H2 levels so if you have some description you'd like for the H1 item and you put it into a separate item it looks a bit ugly, i.e.

1.0 Introduction
1.1 HL002
Some introductory text
1.2 DB Server

versus

1.0 Introduction
Some introductory text
1.2 DB Server

This is something that is currently supported in doorstop , all this change does is correctly apply the attribute to the heading.

jacebrowning:
@michaelnt Thanks for the explanation. I'd be real curious to see a public example of the type of documentation you're producing! 😄 

This change looks harmless because it doesn't change the behavior of one-line headings.

If you can add a test for multi-line headings ([around here](https://github.com/jacebrowning/doorstop/blob/ef309468df5e1f762053b0055daac3569f14f20b/doorstop/core/test/test_publisher.py#L164)) this should be good to go.

PR #222, opened 2017-02-03
Server can listen on a public IP address rather than just 127.0.0.1

michaelnt:
Allow the server to accept connections from a remote host.

I switched to the defaultHelpFormatter to give the hint that the server address can't be a name

### DIFF
--- a/doorstop/server/main.py
+++ b/doorstop/server/main.py
@@ -15,12 +15,20 @@ from doorstop import common, build, publisher
 from doorstop.common import HelpFormatter
 from doorstop.server import utilities
 from doorstop import settings
+from doorstop.core import vcs
 
 log = common.logger(__name__)
 
 app = utilities.StripPathMiddleware(bottle.app())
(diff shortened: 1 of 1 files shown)


doorstop/server/main.py
```diff
@@ -23,6 +24,13 @@
 numbers = defaultdict(int)  # cache of next document numbers
 
 
+class HelpFormatter(argparse.ArgumentDefaultsHelpFormatter):
```
jacebrowning: Is there any reason not to use `ArgumentDefaultsHelpFormatter` as the base for `common.HelpFormatter`?

jacebrowning: Either way, I would like to have all help formatter classes in `common.py`.

michaelnt:
When I started using doorstop I was confused by the searching for
requirements any where in the current repo rather than the current
directory so I think more explicit defaults would be good. So changing the
existing helper is a good idea once all the defaults are also set in the
cli.

PR #227, opened 2017-02-12
Html template and table of contents

michaelnt:
Can now specify a template to use when publishing and optionally insert a table of contents at the top of the file.

An empty `<nav></nav>` is added to the top of the html now when not including a table of contents, which I think is OK.

Should the template be supported for txt and md output?

Should the table of contents be supported for txt and md output?

Could change the help to print out the location of the default html template, to give a clue as to the starting point for customising.

Shall I go ahead and change the template to require bootstrap and jquery and implement the side bar table of contents?

Shall I leave the js and css files on the CDN to save copying them to the publish directory?


jacebrowning:
> Should the template be supported for txt and md output?

Unless you can think of a good use case for templates here, I think the template should only apply when publishing as HTML.

> Should the table of contents be supported for txt and md output?

Probably not.

> Shall I go ahead and change the template to require bootstrap and jquery and implement the side bar table of contents?

Sounds good to me.

> Shall I leave the js and css files on the CDN to save copying them to the publish directory?

I think we want the published HTML to work offline.

michaelnt:
Ok shall I call the option --html-template?

How would you specify the location of the CSS and JavaScript files, perhaps
a --html-includes directory?

jacebrowning:
There was some attempted work a while back to add support for custom CSS: https://github.com/jacebrowning/doorstop/pull/155/files

That approach opted for convention over configuration which I tend to prefer if only a few files are needed to customize a document. Do you think you could find a way to make that work?

michaelnt:
That change embeds the CSS as part of the publishing process which isn't
needed now as you could simply do that in the original template.

I think it is, slightly, better to keep the CSS and JavaScript separate so
they can be shared between documents. A directory would also be helpful for
adding a logo, for instance.

Could copy the contents of an assets folder if located alongside the
template file?

I think an explicit option with the help printing the actual default might
be more discoverable,  although some log.info messages might be enough to
discover the assets mechanism.

jacebrowning:
> I think it is, slightly, better to keep the CSS and JavaScript separate so
they can be shared between documents. A directory would also be helpful for
adding a logo, for instance.

:thumbsup:

> Could copy the contents of an assets folder if located alongside the
template file?

:thumbsup:

> I think an explicit option with the help printing the actual default might
be more discoverable

How about an option that is enabled by default with a sensible default value and updated publishing logic to simply `log.info` if any of the default template files are absent?

michaelnt:
I'm going to add a better template that styles the table of contents as a side bar and includes assets. The default will become to generate a table of contents so the option will become --no-toc?

jacebrowning:
> The default will become to generate a table of contents so the option will become --no-toc

👍 Or just always include the TOC. I think it looks really nice!

michaelnt:
I removed the --no-toc option as you can simply omit the $toc from the html template. I also added a template that lets you keep the previous html format.

jacebrowning:
I think that the absolute path to `doorstop/core/files/sidebar_template.html` should be used as the default for `--template` when it is not provided.

jacebrowning:
`$ make reqs` is not picking up the sidebar template for me. Does `HTMLTEMPLATE` need an absolute path?

michaelnt:
Tried that again and it works for me (on windows).
`doorstop publish --help` shows the absolute path to the template for me.

I don't see how it should fall back to the old template. The side bar will only display once your screen is wide enough. Can you check the output of `doorstop publish --help` and look at the contents of the publish html?

jacebrowning:
Ok, I see what the problem is. `$ make reqs` makes multiple calls to `publish`.

The second call deletes the CSS and JavaScript assets:

```
$ doorstop publish all docs/gen --html
$ doorstop publish all docs/gen --markdown
```

I think we want to support the use case of publishing multiple formats to the same directory. What do you think?

PR #232, opened 2017-02-20
Missing vcs commands raises a DoorstopError

### LINKED ISSUE #230: Missing svn command line tools gives confusing error message
If the svn command line tools are missing doorstop fails to run, even the publish command, and prints a somewhat cryptic file not found traceback.

TortoiseSVN doesn't install the command line svn tools by default.

michaelnt:
Should close #230

### DIFF
--- a/doorstop/core/vcs/test/test_commands.py
+++ b/doorstop/core/vcs/test/test_commands.py
@@ -41,9 +41,11 @@ class BaseTestCase():  # pylint: disable=R0904
         self.wc.commit(self.message)
 
     def test_missing_command(self, mock_call):
-        mock_call.side_effect=FileNotFoundError
+        """Verify that a missing command raises a DoorstopError"""
+        mock_call.side_effect = FileNotFoundError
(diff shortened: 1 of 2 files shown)


doorstop/core/vcs/test/test_commands.py
```diff
@@ -79,44 +85,7 @@ def test_commit(self, mock_call):
 
 
 @patch('subprocess.call')  # pylint: disable=R0904
-class TestMockVCS(BaseTestCase):
-    """Tests for the placeholder VCS plugin."""
```
jacebrowning: Were these tests not working? I wouldn't expect the new logic to affect the "placeholder VCS" because it never actually calls out to the shell.

michaelnt: BaseTestCase.test_missing_command will fail for TestMockVCS and this class never makes any calls. I didn't really see the point of TestMockVCS so I though it made more sense to remove it than to add a work around.

Issue #234, opened 2017-02-24
"doorstop create" deletes entire projects


blastur:
As I tried to create a 2nd doorstop document, doorstop decided to remove my entire project -- not just the doorstop files, but the entire git repo and all other files. Fortunately for me I had pushed my repo elsewhere in the morning, but I lost a days worth of work.

The problem occured when I tried to create a new document that sits above a previously created document in project directory hierarchy. 

Steps to reproduce (using doorstop v1.2b3):
1. Setup the environment
```
$ mkdir asdf
$ cd asdf
$ git init
$ doorstop create ABC reqs/
$ doorstop add ABC -c 5
```
2. Try to create another document that in the asdf dir.
```
$ doorstop create DEF .
building tree...
WARNING: unable to delete: .
ERROR: no parent specified for DEF
```

3. The entire asdf directory is deleted, including the .git repo, the ABC document and all of its items. 

Note that the "doorstop create DEF"-command did not even succeed, doorstop noticed I had forgotten to specify a --parent and printed an error, but it still went ahead and deleted all of my work.

jacebrowning:
Do you know if this bug has always existed?

michaelnt:
I could re-create this. 

The document can't be placed in the tree so it is deleted by calling git rm --force . to the path to the new document being the current directory or root of the repo results in everything being deleted.

It can be fixed by removing the delete command, from [tree.py](https://github.com/jacebrowning/doorstop/blob/develop/doorstop/core/tree.py#L224)  but then a test for the importer fails. The importer [traps an exception to try and detect an unplaced document](https://github.com/jacebrowning/doorstop/blob/develop/doorstop/core/importer.py#L68) and then creates the document a second time which now fails as the document config already exists.

The importer should probably be modified to create the document and then try and place it?

jacebrowning:
> The importer should probably be modified to create the document and then try and place it?

@michaelnt I agree.

@blastur Thanks for the report and sorry about the inconvenience this caused you!

jacebrowning:
This should be fixed in `Doorstop==v1.2.1`.

PR #235, opened 2017-02-25
Deleting a document should not delete the directory unless it is empty

michaelnt:
Deleting a document no longer deletes the folder it contains as that folder could contain assets or have been incorrectly entered and point to some other directory containing non doorstop data.

This solves #234, which I think only affects git.


blastur:
I tested this change, works great!

PR #238, opened 2017-02-26
Add reference documentation for item files

blastur:
This commit adds a new section to the manual that details the
YAML key-value pairs that Doorstop relies on in each item file.


michaelnt:
That looks very helpful.

Would be great if you could explain what derived is used for.

blastur:
Thanks.

About derived, to be honest, I'm not sure :) I suppose it's an attribute you can set when an item is derived from another, like a derived requirement. But I don't know why Doorstop would need to know that, or if it can be used in verification. Maybe Jace can shed some light on it...

jacebrowning:
Here are couple more links on the subject:

- http://www.acqnotes.com/acqnote/tasks/derived-requirements
- http://dwaynephillips.net/advisors/dr.htm

Doorstop needs to know if a requirement is derived because it does not perform upward traceability checks on derived requirements. Derived requirements do not link to any items in a parent document.

PR #250, opened 2017-03-15
Simple pylint fixes

jmacarthur:
These are some of the easiest PyLint warnings to fix; there's still quite a lot left, which I hope to have a look at, but those will need a bit more thought.

### DIFF
--- a/doorstop/core/tree.py
+++ b/doorstop/core/tree.py
@@ -176,9 +176,13 @@ class Tree(BaseValidatable):  # pylint: disable=R0902
             log.info("parent options: {}".format(prefixes))
             raise DoorstopError(msg)
 
+        self.regenerate_children()
+
+    def regenerate_children(self):
         for document in self:
             children = self._get_prefix_of_children(document)
             document.children = children
+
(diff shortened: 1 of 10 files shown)


doorstop/core/tree.py
```diff
@@ -176,9 +176,13 @@ def _place(self, document):
             log.info("parent options: {}".format(prefixes))
             raise DoorstopError(msg)
 
+        self.regenerate_children()
+
+    def regenerate_children(self):
```
jacebrowning: What was the warning that triggered this change? I wonder if this method should be kept private for now (`_regenerate_children()`).

jmacarthur: It was a name clash between the 'document' argument to _place and the 'document' used in the for loop. I couldn't think of a better name for the second use, so I put it in a separate scope. You're right, it should be a private method. I have raised #251 which removes that patch and I'll have another look at this issue tomorrow.

jacebrowning: In the future, you can just amend or revert commits in the same branch. I merged #251. Thanks!

PR #254, opened 2017-03-16
Work around a bug in Python 2.7 & 3.3 which prevents installation

jmacarthur:
See https://bugs.python.org/issue19286. Currently the installer fails
when trying to copy a directory in core/files/.


jmacarthur:
Updated to change _css and _js to doorstop.

PR #256, opened 2017-03-19
Escape quotes when reading the reorder index

### LINKED ISSUE #247: Problem parsing index.yml if comments contain quotes
I tried using the reorder command and it fails on lines like the following

`    - TUT001:  # In this picture !["Pic... `

The comments are now wrapped in quotes and loaded as yaml so the line above becomes

```    
- TUT001:  # In this picture !["Pic... 
    - text : "In this picture !["Pic... "
```

michaelnt:
If the first line of text includes

`["Image"](assets/image.png "Image)`

This can be truncated to

`REQ002 # ["....`

Which can't be parsed as valid yaml. This change escapes the quote character.

This fixes #247

### DIFF
--- a/doorstop/core/document.py
+++ b/doorstop/core/document.py
@@ -443,7 +443,8 @@ class Document(BaseValidatable, BaseFileObject):  # pylint: disable=R0902
         yield "outline:"
         for item in items:
             space = "    " * item.depth
-            comment = item.text.replace('\n', ' ') or item.ref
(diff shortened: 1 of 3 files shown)


michaelnt:
That failure was introduced by a pylint cleanup in acb6e711f23dd4fa1f01d9f9595b8389e9814062.

I've pushed a change to fix the error but it would probably be better to ignore the pylint warning/error in this case (shared function signature)?

PR #262, opened 2017-03-29
WSGI adapter

jmacarthur:
I've made an example WSGI adapter so I can run an example of Doorstop on one of our servers; it might be useful for others.

### DIFF
--- a/docs/reference/webserver.md
+++ b/docs/reference/webserver.md
@@ -3,11 +3,8 @@
 ## Conventional bottle webserver
 
 Doorstop can be run as a standalone web server by running
-`python3 doorstop/server/main.py`.  It will use the current working
-directory as the document source by default. You may need to install
-the [Python Bottle framework](https://bottlepy.org/docs/dev/), which
-can be done with `pip install bottle` or `apt install python-bottle`
-on Debian/Ubuntu systems.
+`doorstop-server`. It will use the current working directory as the
+document source by default.
 
 ## WSGI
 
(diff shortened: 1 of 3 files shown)


docs/reference/webserver.md
```diff
@@ -0,0 +1,32 @@
+# Using the Doorstop web front-end
+
+## Conventional bottle webserver
+
+Doorstop can be run as a standalone web server by running
+`python3 doorstop/server/main.py`.  It will use the current working
```
jacebrowning: This is actually exposed as a command-line client, `doorstop-server`. I think that should be the recommended approach here.

docs/reference/webserver.md
```diff
@@ -0,0 +1,32 @@
...
+directory as the document source by default. You may need to install
+the [Python Bottle framework](https://bottlepy.org/docs/dev/), which
+can be done with `pip install bottle` or `apt install python-bottle`
```
jacebrowning: Bottle will already be installed with `pip install doorstop`.

jacebrowning:
@jmacarthur I tend to prefer keeping the history in tact once a review has started.

FYI: The linter is reporting an unused variable.

jmacarthur:
Hmm, yes, I thought root had more uses in that function but apparently it doesn't. I'll have another go at that.

jmacarthur:
Fixed in this branch; squashed version in jmac/wsgi4 if you'd prefer that.

PR #264, opened 2017-04-05
Jmac/html navigation 2

jmacarthur:
This adds some extra navigation to the web front end and switches over to using Bottle's templates rather than string.template which gives us a bit more flexibility. It's not perfect yet (inline images and the navigation bar are broken) but should be an improvement.

I realise I've been sending in lots of PRs recently; I'm not in any rush to get this in though.


jacebrowning:
👍 on moving to templates!

This seems to break the generated static HTML pages.

jmacarthur:
That's interesting, the tests I did locally (normal `make test` and `doorstop publish all dir`) looked normal to me. What did you see?

jacebrowning:
After running `pipenv run doorstop publish all docs/gen && open docs/gen/REQ.html`:

screen shot 2017-04-05 at 3 43 01 pm

jmacarthur:
Thanks, I'll try and fix this on Friday.

jmacarthur:
Sorry for the delay; pipenv is a bit of a pain to get right. I'm guessing it might be something OSX related from your use of the 'open' command. OSX machines are a bit hard to find here, but I'll keep looking.

jmacarthur:
Ignore that, it's not OSX, I'm just missing a step in my build process.

jmacarthur:
I've had to rewrite the branch history for this, so #268 continues this.

PR #268, opened 2017-04-11
Jmac/html navigation 3

jmacarthur:
Succeeds #264. As per the original, this adds some extra navigation to the web front end and switches over to using Bottle's templates rather than string.template which gives us a bit more flexibility. It's not perfect yet (inline images and the navigation bar are broken) but should be an improvement.

Updated to allow relative URLs for .css & .js files again for static web pages, and adds the previously missing general.css file.


jacebrowning:
I'm going to merge in `master` to get the review app tooling I just added. This will let us browse the generated HTML on PRs.

jmacarthur:
That's a new one on me. Was that with the bottle server or static pages?

jacebrowning:
I see the same behavior with both the server and the static pages.

jmacarthur:
It looks like `$("#main a").attr("target", "parent");` in sidebar.tpl's javascript. I get the same behaviour in the develop branch.

jmacarthur:
Thanks for reviewing!

Issue #260, opened 2017-03-22
Restore pycodestyle rule E741


jacebrowning:
Carryover from https://github.com/jacebrowning/doorstop/pull/258#discussion_r107032612.

PR #270, opened 2017-04-19
Fix pycodestyle E741 and re-enable checks

jmacarthur:


jmacarthur:
Fixes #260.

.pycodestyle.ini
```diff
@@ -6,4 +6,4 @@
 # E711: comparison to None (used to improve test style)
 # E712: comparison to True (used to improve test style)
 # E741 ambiguous variable name
```
jacebrowning: You can remove this line too.

jmacarthur:
Good point! Fixed!

jacebrowning:
I'm going to pin the version of `pydocstyle` in `develop`. The new release has additional checks were not yet using.

PR #275, opened 2017-05-13
Add heading ids when publishing a standalone document

### LINKED ISSUE #257: Assets missing when publishing single document
If I publish all my documents to HTML, e.g

`$ doorstop publish all ./html`

Doorstop will automatically copy assets into the output directory. If I only publish a single document, e.g

`$ doorstop publish REQ ./html/req.html`

.. no assets are copied into the output directory, which will cause the document to render incorrectly and should be considered broken.

Steps to reproduce:
1. Publish a single document with the default template
2. Note neither bootstrap.min.css nor sidebar.css are copied to the destination dir, but referenced from the default template.

michaelnt:
Fixes #257 by adding ids to headings.

Leaves linkify as false so that child/parent links aren't rendered as links to files that may not exist.

Depends on #274 to pass on windows

### DIFF
--- a/doorstop/core/publisher.py
+++ b/doorstop/core/publisher.py
@@ -315,7 +315,7 @@ def _lines_markdown(obj, **kwargs):
                     t=text_lines[0] if text_lines else '')
             else:
                 standard = "{h} {t}".format(h=heading, t=item.text)
-            attr_list = _format_md_attr_list(item, linkify)
+            attr_list = _format_md_attr_list(item, True)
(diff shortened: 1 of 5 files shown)


michaelnt:
This change is passing the tests except for python 3.6, which seems to be a problem with the build environment.

> /home/travis/build/jacebrowning/doorstop/.venv/bin/python: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory

PR #281, opened 2017-09-26
Only ignore relpaths, not entire paths

ojohnny:
There was a problem when my project repo got checked out by a CI runner, where
it got placed into a '.../build/myproject' folder.

In .gitignore, there was a '/BUILD/' ignore path.

As a result, fnmatch.fnmatch(path, '*BUILD*') in BaseWorkingCopy.ignored
returned True (as I understand, it is case insensitive), eventually
resulting in my entire tree being ignored when searching for refs.

There was a 'fix' for this, which don't ignore '*build*' if the
environment variable 'CI' was set.  This didn't match my '*BUILD*'
pattern, so it didn't trigger.  Also, it would actually not ignore a
real 'build/' directory in my repository if there was one, so I consider
this a bug anyway.

### DIFF
--- a/doorstop/core/vcs/base.py
+++ b/doorstop/core/vcs/base.py
@@ -22,7 +22,6 @@ class BaseWorkingCopy(object, metaclass=ABCMeta):
         self.path = path
         self._ignores_cache = None
         self._path_cache = None
-        self._show_ci_warning = True
 
     @staticmethod
     def relpath(path):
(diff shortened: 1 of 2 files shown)


jacebrowning:
Thanks!

ojohnny:
No problem, I believe it is me who should be thanking you for creating a very interesting project!

jacebrowning:
This is released in `doorstop==1.4b1`.

PR #282, opened 2017-09-26
Fix skipping of hidden paths

ojohnny:
1.4b1 introduced a bug where hidden paths in the project root directory
were not skipped (e.g. '.git' directory was not excluded from the paths cache).

### DIFF
--- a/doorstop/core/vcs/base.py
+++ b/doorstop/core/vcs/base.py
@@ -94,7 +94,7 @@ class BaseWorkingCopy(object, metaclass=ABCMeta):
                     if self.ignored(relpath):
                         continue
                     # Skip hidden paths
-                    if os.path.sep + '.' in relpath:
+                    if os.path.sep + '.' in os.path.sep + relpath:
                         continue
                     self._path_cache.append((path, filename, relpath))
         yield from self._path_cache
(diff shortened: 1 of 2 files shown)


doorstop/core/vcs/base.py
```diff
@@ -94,7 +94,7 @@ def paths(self):
                     if self.ignored(relpath):
                         continue
                     # Skip hidden paths
-                    if os.path.sep + '.' in relpath:
+                    if os.path.sep + '.' in os.path.sep + relpath:
```
jacebrowning: Are we sure this line shouldn't simply be `relpath.startswith('.')`? If `relpath` is really a relative path then why are we adding the path separator to it?

jacebrowning: Perhaps there could an additional test to ensure `.git` is not cached?

ojohnny: `relpath.startswith('.')` does not catch `path/to/.hidden/file` though? (I assume that the intension was to exclude all hidden files, not only the ones in the base directory)

I can add a test checking for the existance of `.git` in the path cache, would you recommend to use the paths() generator or should I look in the self._path_cache directly?

ojohnny: And the reason why I was adding the path separator was because `BaseWorkingCopy.relpath()` does not result in a path which starts with the usual './'.  Maybe I should address that instead?

jacebrowning: My understanding is that `os.path.relpath` simply won't start with `/`, not that it has to start with `./`. I think this change looks good!

jacebrowning:
Released in `doorstop==1.4b2`

Issue #31, opened 2014-03-27
Consider adding an optional 'head' attribute to items.


jacebrowning:
The  `head` attribute would be used for publishing headers on **every** Item such as:

```
1.0 Requirements

1.1 REQ001 Start Up

The product shall start up...

1.2 REQ002 Shutdown

The product shall shut down...

1.3 REQ003

The product shall also...
```

Attribute could be called:
- `Item.head`
- `Item.heading`
- `Item.header`

Consider aliasing `Item.text` as `Item.body`.
##

rickeywang:
How would one go about implementing this? Would also like the links to show the heading. 

Instead of a series of cryptic links like `Parent links: REQ-007, REQ-003` , it makes more sense to have a readable description with it. Like `Parent links: REQ-007 Safe Shut Down, REQ-003 Voltage Supply`

jacebrowning:
I haven't thought about the implementation for this in some time, but I would start by looking at how other attributes are managed in the `Item` class.

Once such an attribute is available on items, a modification would be needed in the `publisher` module to include the heading alongside `item.uid` .

neerdoc:
I believe this is already implemented.

PR #285, opened 2017-12-27
Added header attribute to print alongside links and heading.

rickeywang:
For #31 

Introduce a new attribute called `header`, which is displayed alongside the UID and Links for MD/HTML. Gives more context for the Links, and allow for future stylistic changes to the published documents.


doorstop/core/item.py
```diff
@@ -208,6 +212,11 @@ def data(self):
                 value = value.yaml
             elif key == 'text':
                 value = value.yaml
+            elif key == 'header':
+                if hasattr(value, 'yaml'):
```
rickeywang: Handle for cases where header is empty

jacebrowning:
This looks like a good start. I think we want this feature to be truly optional (at least to start) given that nobody else has asked for it thus far. This means that all the existing tests should pass without modification.

After that, you'll want to create an integration test for publishing a document containing items with headings.

jacebrowning:
This is looking pretty good!

Maybe next you'd like to add a CLI option to enable this feature and use this to enable the flag in `cli/utilities/configure_settings`? Or perhaps the `ENABLE_HEADERS` value is simply set via `bool(os.getenv('DOORSTOP_ENABLE_HEADERS'))`.

rickeywang:
The `header` attribute is exported for all other formats as well if the flag is enabled. 

I think it is redundant at this point to have all the tests duplicated for exporting to md, csv, xlsx, etc.. eventually the `header` attribute should become part of the yaml by default. Then just need a flag  to control how it is Published.

jacebrowning:
@rickeywang Thanks for putting in all this work! Would you like to make any more changes before I ship this in a beta release?

rickeywang:
No problem! Looks good now.

jacebrowning:
This is released as `doorstop==1.5b1`.

PR #287, opened 2018-02-16
Put headers on the same line in Markdown/HTML output

ojohnny:
In the newly added 'header' beta feature (see #31), the header was added on a separate line in the Markdown and HTML outputs, like this:

```
## 1.1 REQ002 {#REQ002 }

##Header

Text ...
```

But the 'text' output does something different and puts the header to the right of the UID.  I believe this is the intended format for this feature, so this commit changes the above output to:

```
## 1.1 REQ002 Header {#REQ002 }

Text ...
```

The TOC entry in the HTML output is updated accordingly.

### DIFF
--- a/doorstop/core/publisher.py
+++ b/doorstop/core/publisher.py
@@ -329,7 +329,7 @@ def _lines_markdown(obj, **kwargs):
                 if item.header:
                     uid = '{h} <small>{u}</small>'.format(h=item.header, u=item.uid)
                 else:
-                    uid = '{u} <small>{u}</small>'.format(u=item.uid)
+                    uid = '{u}'.format(u=item.uid)
 
             # Level and UID
             if settings.PUBLISH_BODY_LEVELS:
--- a/doorstop/core/publisher.py
+++ b/doorstop/core/publisher.py
@@ -324,15 +324,19 @@ def _lines_markdown(obj, **kwargs):
             yield from text_lines[1:]
         else:
 
+            uid = item.uid
+            if settings.ENABLE_HEADERS:
+                if item.header:
+                    uid = '{h} <small>{u}</small>'.format(h=item.header, u=item.uid)
+                else:
+                    uid = '{u} <small>{u}</small>'.format(u=item.uid)
+
             # Level and UID
             if settings.PUBLISH_BODY_LEVELS:
                 standard = "{h} {lev} {u}".format(h=heading,
-                                                  lev=level, u=item.uid)
+                                                  lev=level, u=uid)
             else:
-                standard = "{h} {u}".format(h=heading, u=item.uid)
-
-            if item.header:
-                standard = "{s} {h}".format(s=standard, h=item.header)
+                standard = "{h} {u}".format(h=heading, u=uid)
 
             attr_list = _format_md_attr_list(item, True)
             yield standard + attr_list
@@ -464,7 +468,7 @@ def _table_of_contents_md(obj, linkify=None):
             lines = item.text.splitlines()
             heading = lines[0] if lines else ''
         elif item.header:
-            heading = "{u} {h}".format(u=item.uid, h=item.header)
+            heading = "{h}".format(h=item.header)
         else:
             heading = item.uid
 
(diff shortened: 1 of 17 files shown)


jacebrowning:
I'll let the original author take a look.

cc @rickeywang

ojohnny:
The travis build for Python 3.3 failed with the following message:
```
An error occured while installing!
  Could not find a tag or branch '4e7d9fee4080d2e0db67a3e0463be8b196e56a95', assuming commit.
  Could not find a tag or branch 'e0fdd25c214e60bef10fbaa46252f4aaa74de8c2', assuming commit.
isort requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*' but the running Python is 3.3.6
```
It seems like the test environment has changed somehow and python 3.3 is no longer supported?

jacebrowning:
@ojohnny Interesting.

I wonder which dependency is pulling in `isort`? Dropping 3.3 seems reasonable to me.

jacebrowning:
I removed Python 3.3 support in https://github.com/jacebrowning/doorstop/pull/289.

jacebrowning:
@ojohnny Can you enable the beta features on this line so they'll show up in the review app? 

https://github.com/jacebrowning/doorstop/blob/797b92cc3455e1ccb4e0213885c48c5597cf3d21/bin/post_compile#L21

ojohnny:
@jacebrowning I could add a '-b header' to that line if that is what you meant. However, none of the published requirements make any use of the beta features so I don't think you would see any difference.

Do you actually want to see that it does not make a difference or do you want me to add 'header:' fields to all of the requirements as well?

jacebrowning:
@ojohnny Let's add header fields to at least a few of the sample requirements.

ojohnny:
What's the status on this pull request?

doorstop/core/publisher.py
```diff
@@ -330,13 +330,12 @@ def _lines_markdown(obj, **kwargs):
                                                   lev=level, u=item.uid)
             else:
                 standard = "{h} {u}".format(h=heading, u=item.uid)
-            attr_list = _format_md_attr_list(item, True)
-            yield standard + attr_list
 
-            # Item Description
             if item.header:
-                yield ""  # break before text
-                yield heading + item.header
```
rickeywang: Hi Johnny

The original intent for separating the header from rest of information is for the style to look like this: https://imgur.com/qO9TYhZ

I find that having the section, document ID, and the header all in one line and with the same `<h1>` or `<h2>` style quite awful. I've been on and off working on improving the style here: https://github.com/rickeywang/doorstop/tree/bonus_style_points but it is not ready for PR yet. 

For pure text output I agree with you that having them on one line makes most sense.

ojohnny: Hi Rickey,

I still can't see what the benefits are for having the title on two separate lines.  Is this some common practice within requirements documents which I am not aware of?

Specifically, what I don't like about having the two-line title (which is why I created this PR) is:
1. The title in the document has a mismatch from what is seen in the table of contents.  This means you can't visually scan the document, or even use the "find" feature of the browser to find e.g. "2.3 Assets" (because "2.3" and "Assets" are separated)
2. Semantically, having two &lt;h2> tags for one small section makes no sense.  It sort of looks as if there is text missing below the first line.  The style in your upcoming PR seems to correct this somewhat, but I can't really say that I'm a fan of the two-line title. (I find it quite confusing to look at)

What do you think about doing it like this instead:

doorstop-headers

I think it corrects the "awful" problem, while still being coherent with the table of contents.

jacebrowning: I think that looks pretty good with the heading in front of the ID.

rickeywang: Ah yeah this looks nice. And no, I just made up the separate lines style.

jacebrowning: Cool! I think we can merge this and let @rickeywang incorporate the additional styling changes later.

ojohnny:
I I thought it looked more coherent when you have a mix of IDs not using headers and some of them using it, and also that it would be a visual cue to say "You can change this part if you like".  But if you don't like it, I can remove that.

Also, I have no clue why the Python 3.4 check didn't work, and I can't fetch the logs :/

PR #290, opened 2018-03-19
GUI Improvement

elarivie:
Hello @jacebrowning,

This pull requests contains improvement to the GUI.

Some of those fixes are part of the issue jacebrowning/doorstop#202:

- add the ability to click on items in a document **= DONE, Links are now clickable**
- add scrollbars **= Added a Vertical one for the Outline**
- add icon **= Done for Windows & Linux ... (I don't have a Mac)**

Other modifications:

- Fixed a crash occurring when switching document, if the other document had less item than the selected item's index, it would crash the GUI.
- Added a [.editorconfig](http://editorconfig.org/) file
- Replaced the ListView by a TreeView for the Outline (This allows to expand and collapse sections base on item's level)
- Made some read-only field uneditable
- Added some new functionallities/shortcut key:
  - Allows to increase GUI text font size (<kbd>Ctrl</kbd>+<kbd>+</kbd>)
  - Allows to reduce GUI text font size (<kbd>Ctrl</kbd>+<kbd>-</kbd>)
  - Allows to reset GUI text font size (<kbd>Ctrl</kbd>+<kbd>0</kbd>)
  - Allows to delete selected item (<kbd>Delete</kbd>)


doorstop/gui/resources.py
```diff
@@ -0,0 +1,7 @@
...
...lAieVKlz/PwSSRbrGtjeIl/p7DdOwPj4csFWhVA/XRtfVNh1jtckPR9Fd+mPW4ieVmDT8UdZollLYE8SOnRlWMVXWKzQKDujJMbQwix4rqwy+2cGNdVmLX1sqQ/HiRxjLa33R0E9dRbFUvlDv+iVQHgCmMtkRwJFJiekjS3jOu93t5+jxHkaVUAyIdR9kZ/OFDQWik3qBrXrfY2fJaAT6sCQD6NtERyKGAieWCQ65kvZlT5Hmz0pNzECVoVQMYa6uz7tkn6hNx55KMDfF6z3MD+QGckNVH9BrVD0vp+ZZ2kbZZ0AZBAMtMi6UPWKrkuxc/pkPRmkf++W/E2cawHPZJWyXVDbZa0Wm5683luDYA8Gy5piQX0NLpcbivxuetVf91PvXKz0BirAFBTrrJE8kaFg2apTRCfrfFksU9vHasYRjUDUOuJZLEqt7Pte0t8zo9pVQCodo3cgisUduL9V0nvl/RZxduVt5RSeyQdqcJ7c1LSBl0e0F5v//s41QYggeCy85ZI/kXSQ5L+RvF25Y2bQA7l+LuflbRRlwe0V8sNcB+hWgAggcTXJ+lxuS6nhyR9UfF25Y1KIIdz8v32yXVBMQMKAAkk5UTyn3JdW59TvJML85JAegckCVoVAJCRwjbrL2vwQeZ/L/H3H1J6A9t75GZAPSJ3RjozoAAgx4lkVYlg/r8l/t4CVeZgo/4zoOZLGsNPAgDV5wG5Ldv7B/l1Jf7sTFoVAICB5kt62gL/GyX+TMcgrYqVkpbTqgCA+tQgaaFKd2GNkDv06v8kfV1uFfxcuT26AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA6sD/A8eWghJf4lxLAAAAAElFTkSuQmCC"
```
jacebrowning: How do we go about updating these values in the future?

doorstop/gui/main.py
```diff
@@ -99,9 +103,37 @@ def run(args, cwd, error):
...
+        from sys import platform as _platform
+
+        # # Load the icon
+        if _platform in ("linux", "linux2"):
+            # linux
+            from doorstop.gui import resources
+            root.tk.call('wm', 'iconphoto', root._w, tk.PhotoImage(data=resources.b64_doorstopicon_png))
+        elif _platform == "darwin":
+            # MAC OS X
+            pass  # TODO
```
jacebrowning: I'll create an issue for this. I have a Mac, but I'm not sure what is required of icon formats.

doorstop/gui/utilTkinter.py
```diff
@@ -0,0 +1,67 @@
...
+
+import sys
+from unittest.mock import Mock
+try:  # pragma: no cover (manual test)
+    import tkinter as tk
+    from tkinter import ttk
+except ImportError as _exc:  # pragma: no cover (manual test)
+    sys.stderr.write("WARNING: {}\n".format(_exc))
+    tk = Mock()
+    ttk = Mock()
```
jacebrowning: We may not need this pattern anymore. I think I added it for testing, but I'm not sure how much value we can get out of unit testing this GUI.

elarivie: Ok, I simply copy pasted the import logic from the other file.  We can keep it like that for now, at least the import strategy is the same everywhere.

doorstop/gui/main.py
elarivie: Base on this [stackoverflow question](https://stackoverflow.com/questions/33134594/set-tkinter-python-application-icon-in-mac-os-x?rq=1), it seems to require the image to be in ".icns" format

doorstop/gui/resources.py
elarivie: Those values are [Base64](https://en.wikipedia.org/wiki/Base64) encoded data of the DoorStop logo.

For the b64_doorstopicon_png, I simply took [doorstop/docs/images/logo-black-white.png](https://github.com/jacebrowning/doorstop/blob/develop/docs/images/logo-black-white.png) which I encoded using:

    /usr/bin/base64 logo-black-white.png

For the b64_doorstopicon_ico, I took [doorstop/docs/images/logo-black-white.png](https://github.com/jacebrowning/doorstop/blob/develop/docs/images/logo-black-white.png) which I converted to .ico and then base64 encoded it (I don't remember which free online image format converter website I used)

**Note:** Embedding the image in base64 directly in a python source file has the big advantage that it can be easily accessed using imports instead of trying to guess the file path to get the image at runtime. + I grouped them in the same file so that the blobs of text does not slowdown text editor and so that it does not pollute files where python development is done.

In the future, unless the DoorStop icon is modified, there is not much reason to edit those string.  This file could be use to embed other image if needed.

**Note:** Those base64 string can easily be decoded to recover the original image.

Issue #298, opened 2018-05-15
Update documentation for latest GUI updates


jacebrowning:
https://github.com/jacebrowning/doorstop/tree/develop/docs/gui

PR #301, opened 2018-05-30
GUI Improvement Part 2

elarivie:
Hello @jacebrowning,

To follow on pull request #290, this pull request contains other improvements mainly related to the GUI.

I completely restructured the GUI, the widgets and the model are now correctly decoupled.  I used an architecture similar to React&Redux (Store, State, Action, Reducer).

I also added some missing features which were previously not available from the GUI.

**From the GUI it is now possible to:**
- Open/Close project
- Save (or Exit without saving)
- Export
- Import
- Publish
- Activate/Deactivate items (it was previously not possible since deactivated items were disappearing from the list, it was therefore not possible to re-activate items)
- I added menu items + keyboard short-cut
- I removed the "Items" read-only text field (it was pretty much useless to have ALL the document text in one big read-only text field)... Instead if someone wants to quickly preview the document, it is now possible to do: "File" > "Publish Tree (Preview 5 minutes)"
    - **This automatically:**

    1. Publish the tree in a temp folder
    2. Open a web browser to the Index.html
    3. -- The user can browse the documents/items --
    4. 5 minutes later, it auto delete the temp folder.

**From the CLI it is now possible to:**
- List items

**Other stuff:**
- I started to add [Mypy](http://mypy-lang.org/) annotation

**Note:**
While doing those changes I found a bug in doorstop... when AutoSave is False... when adding/removing items... the files instantatenously get created/deleted on disk without waiting for the explicit save. (_This pull request does not fix this bug_)

**Note:**
I am tracking possible future improvement to the GUI in a TODO file that I commited within the doorstop/gui folder.

**Note:**
I saw that you created #298 to document the GUI, I was thinking about creating a static single html page describing the GUI which could be available from GitHub by browsing the code + create a F1/HelpMenuItem from within the GUI to open the same html page.


Makefile
```diff
@@ -64,7 +64,7 @@ check: pylint pycodestyle pydocstyle ## Run linters and static analysis
 
 .PHONY: pylint
 pylint: install
-	$(PYLINT) $(PACKAGES) $(CONFIG) --rcfile=.pylint.ini
+	$(PYLINT) $(PACKAGES) $(CONFIG) --rcfile=.pylintrc
```
jacebrowning: I prefer the `.pylint.ini` filename so that everyone's editor gets syntax highlighting by default.

elarivie:
Here is a detail of the doorstop core changes which were done to support the GUI improvement:

**Core change # 1**

doorstop/core/__init__.py

**Description**
- Added the boolean parameter "only_active" which default to True (to match previous behavior) to the method "get_issues" so that it is now possible to get ALL issues including those which are inactive.

**Justification**
This allows in the GUI to deactivate AND re-activate items.  Without this, the deactivated item could not be shown in the GUI.

**Core change # 2**

doorstop/core/base.py

**Description**
- Removed the shared static global boolean value "auto = True"
- Added "is_auto_save" as internal property or parameter to bring this flag where the orignal "auto" flag was used.

**Justification**
- Doing so allows to have full control about being able to instantiate some items as being auto saved and other not.
- Not relying on a global shared setting is also a better design.


The above two changes also lead to update other files and unit test to match the change.

.pylint.ini
```diff
@@ -123,14 +123,23 @@ disable=
...
+    unnecessary-lambda,
+    using-constant-test,
+    C0122,
+    C0302,
+    C0321,
+    R0201,
+    W0108
+    W0125,
+    W0201,
+    W0613
```
jacebrowning: Please use the name of the rule rather than the ID.

With CII, we're need to be deliberate here: https://bestpractices.coreinfrastructure.org/en/projects/754

doorstop/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 from doorstop.common import DoorstopError, DoorstopWarning, DoorstopInfo
 from doorstop.core import Item, Document, Tree
-from doorstop.core import build, find_document, find_item
```
jacebrowning: Some people might depend on these APIs, so if required, this will be a breaking change.

doorstop/cli/commands.py
```diff
@@ -411,7 +449,7 @@ def run_import(args, cwd, error, catch=True, _tree=None):
         utilities.show("imported document: {} ({})".format(document.prefix,
                                                            document.relpath))
     else:
-        assert item
+        assert item, "Should have an item."
```
jacebrowning: Thanks for this! I agree that all asserts should have messages.

akuker:
Dang! I wish I had seen this earlier. I've been working on a C#/.NET version of the GUI with a look/feel similar-ish to the commercial tool. 
https://github.com/akuker/doorstop.net

high level view
document viewer
item editor2
item editor

doorstop/core/item.py
```diff
@@ -600,11 +597,6 @@ def get_issues(self, skip=None, document_hook=None, item_hook=None):  # pylint:
                 else:
                     yield DoorstopWarning("unreviewed changes")
 
-        # Reformat the file
-        if settings.REFORMAT:
```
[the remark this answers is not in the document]
elarivie: About this specific change

Here is the original code:
```

    def get_issues(self, skip=None, document_hook=None, item_hook=None):  # pylint: disable=unused-argument
        """Yield all the item's issues.
        ...

        # Reformat the file
        if settings.REFORMAT:
            log.debug("reformatting item %s...", self)
            self.save()

```

By looking at the code, it didn't make sens that a getter would silently trigger an item save to disk, so I removed this "if" block.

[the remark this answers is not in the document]
jacebrowning: Modifications can occur on-disk, so I think this block was there to apply formatting to changes made directly to the YAML files.

doorstop/gui/reducer.py
```diff
@@ -51,7 +51,7 @@ class Reducer(object):
     def __init__(self) -> None:
         pass
 
-    def reduce(self, state: State, action: Action) -> State:
+    def reduce(self, state: State, action: Action) -> State:  # pylint: disable=R0201
```
jacebrowning: Same here. Prefer the rule name so that no one has to look it up. 😄

doorstop/core/base.py
```diff
@@ -183,14 +183,13 @@ class BaseFileObject(object, metaclass=abc.ABCMeta):
...
-    auto = True  # set to False to delay automatic save until explicit save
-
-    def __init__(self):
+    def __init__(self, is_auto_save):
         self.path = None
         self.root = None
         self._data = {}
         self._exists = True
         self._loaded = False
+        self.is_auto_save = is_auto_save
```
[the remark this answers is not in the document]
elarivie: Doing this search and replace could possibly lead to some ambiguity

For example in the file core/base.py:
```python3
def auto_save(func):
    """Call self.save() after execution."""
    @functools.wraps(func)
    def wrapped(self, *args, **kwargs):
        result = func(self, *args, **kwargs)
        if self.is_auto_save:
            self.save()
        return result
    return wrapped
```

Currently the different names are:
* **save** ➔ The method to actually save to disk
* **is_auto_save** ➔ A boolean value
* **auto_save** ➔ A method wrapper handling auto saving

I think it is better to keep it like that.

doorstop/__init__.py
elarivie: I really doubt that any body would be relying on those methods.. From what I saw they were used by unit test only... they were providing a static and shared tree to all unit tests... With little change they were then no more needed so I removed them.

**Pro:**
* No static shared variables
* Do not keep code which is not needed

doorstop/core/base.py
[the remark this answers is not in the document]
jacebrowning: I see. Then maybe `is_auto_save` could be `should_auto_save`?

Pipfile
```diff
@@ -41,3 +41,6 @@ twine = "*"
 sniffer = "*"
 pync = { version = "<2.0", sys_platform = "== 'darwin'" }
 MacFSEvents = { version = "*", sys_platform = "== 'darwin'" }
+
+# This is needed for Python 3.4
+typing = "*"
```
jacebrowning: 👍

doorstop/__init__.py
jacebrowning: Regardless, they were part of the public API. It's not big deal, but this change will have to be part of the next major release.

wrc-jasonw:
Hello elarivie and jacebrowning!

I might be willing to provide manpower to resolve some of the issues here.

Is there anything I can do help get the GUI improvements merged into either a branch or into mainline?

Thanks!
-Jason White

jacebrowning:
For this PR, I am a bit hesitant about passing `should_auto_save` around multiple layers of the API.

Is there any way to accomplish these improvements to the GUI without so many changes to the core library?

RomainTT:
> Dang! I wish I had seen this earlier. I've been working on a C#/.NET version of the GUI with a look/feel similar-ish to the DOORS. I should have spent my time on the python GUI.
> https://github.com/akuker/doorstop.net
> ...

That GUI is soooooooooooooooo nice. It is exactly what I am looking for. Is there any hope this PR makes the GUI a little bit like this ?

JbalTero:
@elarivie 
after `python setup.py clean --all install` and `doorstop-gui` I get this error

```
λ doorstop-gui
Traceback (most recent call last):
  File "C:\Users\jbalt\AppData\Local\Programs\Python\Python37\Scripts\doorstop-gui-script.py", line 11, in <module>
    load_entry_point('Doorstop==1.6b1', 'console_scripts', 'doorstop-gui')()
  File "C:\Users\jbalt\AppData\Local\Programs\Python\Python37\lib\site-packages\pkg_resources\__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "C:\Users\jbalt\AppData\Local\Programs\Python\Python37\lib\site-packages\pkg_resources\__init__.py", line 2793, in load_entry_point
    return ep.load()
  File "C:\Users\jbalt\AppData\Local\Programs\Python\Python37\lib\site-packages\pkg_resources\__init__.py", line 2411, in load
    return self.resolve()
  File "C:\Users\jbalt\AppData\Local\Programs\Python\Python37\lib\site-packages\pkg_resources\__init__.py", line 2417, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "c:\users\jbalt\appdata\local\programs\python\python37\lib\site-packages\doorstop-1.6b1-py3.7.egg\doorstop\gui\main.py", line 25, in <module>
    from typing import Any
  File "c:\users\jbalt\appdata\local\programs\python\python37\lib\site-packages\typing-3.6.6-py3.7.egg\typing.py", line 1356, in <module>
    class Callable(extra=collections_abc.Callable, metaclass=CallableMeta):
  File "c:\users\jbalt\appdata\local\programs\python\python37\lib\site-packages\typing-3.6.6-py3.7.egg\typing.py", line 1004, in __new__
    self._abc_registry = extra._abc_registry
AttributeError: type object 'Callable' has no attribute '_abc_registry'
```

elarivie:
@JbalTero I googled the stack trace that you have and found  [another project issue with the exact same error and some explanation](https://github.com/easybuilders/easybuild-easyconfigs/pull/7917):

Basically this is due to the package **typing**

 which I added in the dependency list:

image

So effectively when I wrote this pull request the package typing was needed.

But time has pass and python 3.7 is now out and the package typing is no more needed.
.

**See below a screen shot of the  explanation**
image
.
.
💡 **The solution is therefore to simply remove the package typing from the dependency list**

@JbalTero I don't know but maybe you also have to locally uninstall the no more needed package typing from your python installation

JbalTero:
@elarivie ok I'll give it a try.

jacebrowning:
I'm going to close this due to age and the volume of conflicts that have arisen since this was originally created. Feel free to open new PRs incorporating these ideas.

Issue #277, opened 2017-05-31
When publishing markdown test in headings is duplicated


michaelnt:
If a heading item contains an additional paragraph of text this will be duplicated when exporting to markdown.

The html output works fine.

guille-r:
It seems there's an error in formatting the heading when exporting to markdown with command option "--no-levels all". 
HTML is also affected as the publisher export to markdown first and then to html.

to reproduce :

`$ doorstop create REQ ./reqs` 

`$ doorstop add REQ`

REQ001.yaml 
```
active: true
derived: false
level: 1.0
links: []
normative: false
ref: ''
reviewed: null
text: |
  Doorstop **shall** output heading

  And body only once.
```
then publish with :
`doorstop publish --no-levels  all REQ reqs.md`

Ouput is :
```
# Doorstop **shall** output heading

And body only once. {#REQ001 }

And body only once.
```

It should be :
```
# Doorstop **shall** output heading {#REQ001 }

And body only once.
```

RomainTT:
@jacebrowning Maybe you can close this issue since it has been fixed ? :)

PR #302, opened 2018-08-10
When publishing markdown test in headings is duplicated #277

guille-r:
In markdown Every text line were printed into the heading when disabling heading level with command line "--no-level all"

see issue:  #277


guille-r:
Ok I'll do.

The only thing is that I have difficulties setting up the development environment on my Win10 platform (the makefile is not very windows friendly :p, even with MinGW). Not sure if I'll be abble to run tests on my own before publishing. 

I'll give a try with Cygwin.

PR #316, opened 2018-11-01
Support $EDITOR from environment. Make --tool required if not available.

aars:
How's this?

PS: Lol. My shell expanded `$EDITOR` in the commit message.

### DIFF
--- a/doorstop/cli/main.py
+++ b/doorstop/cli/main.py
@@ -14,6 +14,7 @@ log = common.logger(__name__)
 
 EDITOR = os.environ.get('EDITOR')
 
+
 def main(args=None):  # pylint: disable=R0915
     """Process command-line arguments and run the program."""
     from doorstop import CLI, VERSION, DESCRIPTION
--- a/doorstop/cli/main.py
+++ b/doorstop/cli/main.py
@@ -12,6 +12,7 @@ from doorstop.core import publisher, vcs, document
 
 log = common.logger(__name__)
 
+EDITOR = os.environ.get('EDITOR')
 
 def main(args=None):  # pylint: disable=R0915
     """Process command-line arguments and run the program."""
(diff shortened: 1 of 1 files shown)


doorstop/cli/main.py
```diff
@@ -180,8 +181,9 @@ def _edit(subs, shared):
                        help="edit document as exported XLSX")
     required = sub.add_argument_group('required arguments')
     required.add_argument('-T', '--tool', metavar='PROGRAM',
-                          help="text editor to open the document item",
-                          required=True)
+                          default=EDITOR,
+                          required=not bool(EDITOR),
```
jacebrowning: _Does_ this ever need to be required?

If I unset `$EDITOR` and run `$ doorstop reorder REQ` it will still open up the YAML file in my default text editor. I think we should have the same fallback behavior for editing items.

aars: On osx I get 
```
$ pipenv run doorstop edit REQ001
No application knows how to open /Users/aaronh/src/doorstop/reqs/REQ001.yml.
ERROR: no default editor for: /Users/aaronh/src/doorstop/reqs/REQ001.yml
```

jacebrowning: Ah, OK. I must have set a default editor for `.yml` on my system.

In that case, I think same `required=not bool(EDITOR)` logic should be added for `$ doorstop reorder`. I assume the reorder command fails on your machine when you unset `$EDITOR`?

aars: @jacebrowning No. Reorder simply goes through with auto-reordering.

```
export EDITOR=
pipenv run doorstop reorder REQ
building tree...
reorder from 'reqs/index.yml'? [y/n] y
reordering document REQ...
reordered document: REQ
```

I think your initial suggestion would be best, to simply remove `required` for the editor. The Operating System message about no application available to open the specific file is fine.

I have updated the PR to reflect this.

jacebrowning:
This looks good to me. The last thing to go is a bit of code formatting, see the `$ make check` output.

PR #317, opened 2018-11-04
links to markdown files should end with .md

### LINKED ISSUE #314: Published markdown has links to HTML files instead of markdown files
Running `doorstop publish -m all ./output` generates markdown files that contain links to HTML files rather than markdown files:

`*Child links:* [TST002](TST.html#TST002)`

Probably a minor issue, might open a PR myself later.

Korijn:
fixes #314

### DIFF
--- a/doorstop/core/publisher.py
+++ b/doorstop/core/publisher.py
@@ -547,7 +547,7 @@ def check(ext):
     :return: lines generator if available
 
     """
-    exts = ', '.join(ext for ext in FORMAT_LINES)
+    exts = ', '.join(e for e in FORMAT_LINES)
     msg = "unknown publish format: {} (options: {})".format(ext or None, exts)
     exc = DoorstopError(msg)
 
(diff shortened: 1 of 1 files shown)


Korijn:
Turns out the general approach to publishing has to be revised (somewhat) to makes this work. The changes now generate correct file extensions in links, but publishing to HTML first generates markdown and then converts using the `markdown` package. Of course, this doesn't convert file extensions in links...

jacebrowning:
Thanks! There should be tests to update along with this change.

Korijn:
Well, actually, the tests are good... They are failing not because there is a bug, but because the implementation doesn't really support linking properly. I need to do some refactoring later.

sebhub:
I am not sure what needs to be done here.

Korijn:
Well unfortunately this PR should not be merged. I thought it was an easy fix, but fixing links for markdown breaks them for other formats like html. The underlying design is flawed and needs more attention than I have given it here, in the form of refactoring.

PR #322, opened 2019-03-22
upgrade to PyYAML 5.1

### LINKED ISSUE #321: Cannot be installed with PyYAML 5.1
I am working on an environment where already a version 5.1 of PyYAML is required. This update has been recently release due to security concerns. (See https://msg.pyyaml.org/load for reference)

As far as I see it, allowing doorstop to be used with this version of PyYAML should not be a big change. (I will prepare a Pull Request)

patkan:
* allow PyYAML version 5
* add Loader-argument to yaml.load
* use same Loader as before
* more information at https://msg.pyyaml.org/load
* fixes #321

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setuptools.setup(
     ],
 
     install_requires=[
-        "PyYAML >= 3.10, < 6",
+        "PyYAML >= 5.1, < 6",
         "Markdown >= 2, < 3",
         "openpyxl >= 2.1, < 2.2",
         "bottle == 0.12.13",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -80,7 +80,7 @@ def build_description():
     ],
 
     install_requires=[
-        "PyYAML >= 3.10, < 4",
+        "PyYAML >= 3.10, < 6",
```
jacebrowning: Thanks! Let's require at least 5.1 here.

patkan: OK, if we require at least 5.1 (added in 
9b8ba29), we could also use `FullLoader` (added in 
05eba75).

PR #326, opened 2019-04-11
Fix #24

### LINKED ISSUE #24: `doorstop edit <ID>` should only open the text by default
New behavior:

```
doorstop edit <ID>  # opens only Item.text in an editor

doorstop edit --all <ID>  # opens Item.path in an editor (current behavior)
```
##

RomainTT:
Fix issue #24. Add --all option to `doorstop edit`.

Now, `doorstep edit ITEM` only the text part of the item by default.
To edit the whole item with all of its attributes, one must add the
--all or -a option to the command.

To edit the text part only, a temporary file is created to be edited
by the user, and then read by doorstop to update the actual item file.

### DIFF
--- a/doorstop/cli/commands.py
+++ b/doorstop/cli/commands.py
@@ -188,7 +188,7 @@ def run_edit(args, cwd, error, catch=True):
 
         # edit it
         if item:
-            item.edit(tool=args.tool)
+            item.edit(tool=args.tool, edit_all=args.all)
         else:
             _export_import(args, cwd, error, document, ext)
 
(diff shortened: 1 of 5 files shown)


RomainTT:
I fixed the test for `doorstop edit ITEM --all` but I could not find a good solution to test the default behavior. I am not used to `unittest` and I don’t know what to do. Maybe you could help me. Or we can say it is not so important for this PR.

RomainTT:
Hi @jacebrowning 
Any comment on this PR ?

PR #328, opened 2019-04-23
Fix #304

### LINKED ISSUE #304: first item in a level (level: 1.0) gets formatted differently from the rest (level: 1.1, 1.2, etc.) when `door publish`
Is this intentional, that when `door add REQ` three times, then `door publish REQ doc.html` or `door publish REQ doc.md`

The first item is one level up than the rest. 

It can be traced back to the `def _format_level(level):` function inside `publisher.py` where

    def _format_level(level):
    """Convert a level to a string and keep zeros if not a top level."""
    text = str(level)
    if text.endswith('.0') and len(text) > 3:
        text = text[:-2]
    return text

So `level: 1.0` the ending zero was thrown out. 

Is there a reason for that? Prettier document or a traditional style?

RomainTT:
Update description of headings to avoid confusion.

### DIFF
--- a/docs/reference/items.md
+++ b/docs/reference/items.md
@@ -1,7 +1,7 @@
 # Overview
 
 Doorstop items are files formatted using YAML. When a new item is added using
-```doorstop add```, Doorstop will create a YAML file and populate it with all
+`doorstop add`, Doorstop will create a YAML file and populate it with all
 required attributes (key-value pairs).
 
 Example item:
@@ -68,7 +68,7 @@ item has not yet been reviewed. Doorstop will use this to detect unreviewed
 changes to an item by comparing the current item fingerprint to the last
 reviewed fingerprint.
 
-You should not calculate this value manually, use ```doorstop review```.
+You should not calculate this value manually, use `doorstop review`.
 
 
 ### links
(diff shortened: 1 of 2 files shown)


RomainTT:
Fix #304 (just a comment for github automation)

docs/reference/items.md
```diff
@@ -164,7 +164,7 @@ text: |
 When this item is published, Doorstop will create a new heading with the text
 "1.1.0 This is the heading" and put the remaining text into its body.
 
-#### Example: Normative item
+#### Example: Normative heading
```
jacebrowning: I believe this will only be rendered as a heading if there are items nested under it, correct?

RomainTT:
```shell
doorstop create SRS srs
doorstop add SRS
doorstop publish SRS one_item.md
doorstop add SRS
doorstop add SRS
doorstop publish SRS three_items.md
```

Result in *one_item.md*:

```
# 1.0 SRS001 {#SRS001 }
```

Result in *three_items.md*:

```
# 1.0 SRS001 {#SRS001 }

## 1.1 SRS002 {#SRS002 }

## 1.2 SRS003 {#SRS003 }

```

It seems it is always rendered as a heading. Or am I missing something?

jacebrowning: Well, they're all technically "headings" in Markdown. 😄 

The more that I think about this, I think that "Normative heading" will cause more confusion. I'm inclined to close #304 as "working as intended". I would appreciate a cleanup PR that fixes the triple backticks.

PR #331, opened 2019-05-31
For publish cmd, add --no-assets args so that it is possible to exclude copying default doorstop /core/files/assets directory.

eudora-fabia:
This is useful to be able to customize assets directory.

### DIFF
--- a/doorstop/cli/commands.py
+++ b/doorstop/cli/commands.py
@@ -498,6 +498,9 @@ def run_publish(args, cwd, error, catch=True):
     if args.width:
         kwargs['width'] = args.width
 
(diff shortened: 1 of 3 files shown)


michaelnt:
If I include pictures I place them in the assets folder so they will still be visible in the published documents so not copying the assets folder could be a problem.

You might want to do this so you can edit the html, but there is a template option you can use. You can copy doorstop/views/sidebar.tpl to some location and edit the path to the css files to be absolute.

Does that solve the problem this pull request was addressing?

jacebrowning:
I'm going to close this due to inactivity. Feel free to open a new PR!

PR #333, opened 2019-06-01
Expose document object to support better templating.

JbalTero:
When the document object is exposed for templating, you will be able to utilize doc obj attributes such as the current doc name, config, children, items, etc.

### DIFF
--- a/doorstop/core/publisher.py
+++ b/doorstop/core/publisher.py
@@ -524,7 +524,7 @@ def _lines_html(obj, linkify=False, extensions=EXTENSIONS,
                                                      '..', 'views'))
             if 'baseurl' not in bottle.SimpleTemplate.defaults:
                 bottle.SimpleTemplate.defaults['baseurl'] = ''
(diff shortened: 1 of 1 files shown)


doorstop/core/publisher.py
```diff
@@ -524,7 +524,7 @@ def _lines_html(obj, linkify=False, extensions=EXTENSIONS,
                                                      '..', 'views'))
             if 'baseurl' not in bottle.SimpleTemplate.defaults:
                 bottle.SimpleTemplate.defaults['baseurl'] = ''
-            html = bottle_template(template, body=body, toc=toc_html, parent=obj.parent)
+            html = bottle_template(template, body=body, toc=toc_html, parent=obj.parent, obj=obj)
```
jacebrowning: Maybe `obj=` should be `document=` and `bottle_template()` should be rewritten to use `document.parent`?

JbalTero: I refactored it to `document=`
I didn't remove `parent=obj.parent` for backwards compatibility.

Issue #332, opened 2019-05-31
Feature Proposal: Math Expressions in LaTeX syntax


tangoalx:
It would be great, if it is possible to write math expressions in LaTeX syntax in the `text` attribute of an item and also if this expression is rendered  in HTML.

- I propose the usage of the package `python-markdown-math` for parsing of LaTeX code in `text` attribute. That package extends the already used `Markdown` package. 
- Render in HTML can be solved by the wide-spread javascript library MathJax.

Note: MathJax is quite a big javascript collection which should be avoided to be added directly to the repository of doorstop.

jacebrowning:
Added in https://github.com/jacebrowning/doorstop/pull/334

PR #334, opened 2019-06-01
Integrate LaTeX in text attribute and render in HTML

tangoalx:
See #332.

Notes:
- The MathJax stuff is a little messy as you may see in setup.py. I would appreciate, if you find a cleaner way.
- I didn't want to add MathJax files to the git repository so it is downloaded during installation. Maybe also a way for node.js, etc. and future libraries.
- I extended the documentation with an example how to use the LaTeX syntax in item.yaml


jacebrowning:
How big are the MathJax files? Could we delay download until publication?

Or even better, could we instead link to `MathJax.js` on an external CDN to avoid needed to download the files locally?

tangoalx:
MathJax is over 30 MB. Yes, we can do either ways. I also prefer just linking against a CDN. I will change it that way. 

The reason why I downloaded it is that:
- we don't depent on the availibility of an external service 
- the publication is self-contained

However, the availibility should be stable nowadays. And maybe we can achieve self-containability in another way.

tangoalx:
Wow, finally got all the tests passed. 

As discussed, we link now MathJax from a public cdn (cdnjs.cloudfare.com).

I had to correct the "publish" test cases, which failed due to new code in template.html.

Furthermore, I also added some Math Expressions in a requirement (see [REQ2-001.yml)](https://github.com/jacebrowning/doorstop/pull/334/commits/42f93dd829d0a2f424cb7ab5f79aff1904b539ed#diff-d2d4797676220e5c98e0d78412156e68/doorstop/core/tests/files/REQ2-001.yml) and adapted the expected results in published txt, md, html, csv, tsv, excel accordingly. In HTML it looks like this (having also context menu provided by MathJax with all its settings and features like languages, show latex command, accessibility, etc. ):


latex-integration


I would be glad, if you integrate these changes and close this pull request. Looking forward to dive deeper into doorstop. It 's concept is exactly what I have searched for ages.

Issue #335, opened 2019-06-02
Feature Proposal: Render PlantUML diagrams


tangoalx:
Imagine an item `OBL010.yml`:
```
    active: true
    derived: false
    level: 1.2.3.1
    links: []
    normative: true
    ref: ''
    reviewed: null
    text: |
      Cyclic Program Activation

      ```plantuml
      @startuml
      [*] -right-> idle
      idle -right-> fetch : new cycle event
      fetch -down-> process : queue is empty
      process -left-> output : precess finished
      output -up-> idle : output stable
      @enduml
      ```    
```
This shall be rendered as follows:
![plantuml-integration-proposal](https://user-images.githubusercontent.com/46009720/58762806-da22bc00-8553-11e9-9a1e-73ef850cd5d7.png)

This can be achieved by using [PlantUML](http://plantuml.com/de/index), a textual description format for diverse diagram types like `Sequence Diagram`, `Class Diagram`, `Activity Diagram`, `State Chart Diagram` and many more. 

Fortunately, we can integrate this feature with the python package [plantuml](https://pypi.org/project/plantuml/), which works together with a public [plantuml server](http://www.plantuml.com/plantuml/img/). That plantuml server processes the textual input and renders it into a downloadable image, which we can store in the `assets` folder of a publication (don't rely that the image will be available on the server forever).

Note also, that the python package [plantuml](https://pypi.org/project/plantuml/) already allows to redirect your plantuml code to an own plantuml server and let it render your diagrams instead of rendering it on a public server. This could be interesting for companies. Therefore, the server address shall optionally be overridden by an argument to doorstop.

Of course it would be possible to keep the `plantuml code` apart of the doorstop items and just link the image into the item. However, this does not feel natural and has some drawbacks:
- as author, you need to keep the item and the plantuml code in sync, which is more complicated, because they are not in the same file
- changing the image would not lead to `suspected` status automatically
- if you want to get the `suspected` status when changing the image, then you need to change the link address, so the file name of your generated plantuml image must always change. This seems fragile and may lead to invalid links.

JustinW80:
I've been pitching Doorstop to my team for requirements while also boosting PlantUML alongside in design description documents. Integrating PlantUML in Doorstop's text field would be a big win for our use case.

tangoalx:
Glad to hear that this feature will serve somebody else, too!

I've made a draft version here: 80206fe1431165260a7b62f058d1d2416914e2dd

[...]

PR #338, opened 2019-06-11
Render PlantUML diagrams in HTML export

tangoalx:
According to #335.

By default: accessing http://www.plantuml.com/plantuml to generate plantuml diagrams.

Images are integrated in HTML instead of stored in asset folder.

REQ002.yml extended to show exemplary usage in tests. Updated tests accordingly.


doorstop/core/publisher.py
```diff
@@ -18,7 +19,13 @@
     'markdown.extensions.extra',
     'markdown.extensions.sane_lists',
     'mdx_outline',
-    'mdx_math'
+    'mdx_math',
+    PlantUMLMarkdownExtension(server='http://www.plantuml.com/plantuml',
+                              cachedir='/tmp',
```
jacebrowning: Will this work on Windows? Perhaps `tempfile.gettempdir()` will be more portable.

tangoalx: Unfortunately, I don't have a windows system available to test it. I can change it anyway to see if it still works.

jacebrowning:
Thanks for putting this together! Please add an example diagram in [`TUT017`](https://github.com/jacebrowning/doorstop/blob/develop/reqs/tutorial/TUT017.yml) (or make a new item) to show off this feature.

tangoalx:
@jacebrowning: how is the tutorial updated on github.io? Do you do it if a new release is published?

jacebrowning:
@tangoalx This was supposed to be happening automatically, but it looks like my token expired.

Your changes are now live: http://jacebrowning.github.io/doorstop/TUT.html#TUT017

PR #340, opened 2019-06-25
add extended reviewed attributes

TingIsGitting:
First added Add MockSimpleDocument.
Then enable extended reviewed attributes (mentioned in https://github.com/jacebrowning/doorstop/issues/337).
But still need to keep `if self.document:` in `stamp()` of `item.py`.
BTW, in `get_issues()` of `item.py`, there are two places using `if self.document` and `self.document` is passed as parameter. If we remove the `if self.document`, there will be bugs for many unit tests which don't have a document.


sebhub:
We will try to refactor the commits and make them easier to review and integrate.

PR #343, opened 2019-07-01
Prevent "doorstop link" to add a self references or cyclic dependency

sebhub:


sebhub:
Sorry for the forced pushes. I had to learn that a "make ci" does more checks than a "make test", e.g. style checks. Now I am done for the first review.

sebhub:
I just noticed that this implementation has a weak point. In case there is already a cycle in the graph, then the check_for_cyclic_dependency() will not terminate.

sebhub:
I had to fix a comment in the third commit. Everything is now ready for review.

PR #353, opened 2019-07-04
Improve styling for HTML Export

rickeywang:
Changed the index page formatting to use sans-serif and default bootstrap styling so it is consistent with other HTML pages that were generated.

new index

* Made some fonts smaller, added `<hr>` between items for better visibility.
* Fixed a bug where if there is a subheading too many levels down it runs out of levels and becomes `<h1>` 
* Fixed overlapping nav menu for certain page widths
* Simplified styling for Parent and Child links
* Adding a link to go back to Index in the left nav menu

doc improvements

* If the `-b header` flag is used for the Header beta feature, then the contents of each item's header will be displayed instead of the item's ID

header-name-in-title


rickeywang:
Would like to get some feedback from everyone with regards to the styling first. 

Then I can update the unit tests to reflect the changes.

doorstop/core/publisher.py
```diff
@@ -522,6 +537,7 @@ def _lines_html(obj, linkify=False, extensions=EXTENSIONS,
 
     if toc:
         toc_md = _table_of_contents_md(obj, True)
+        toc_md = '[🡨 Index](index.html) \n\n ' + toc_md  # Add a link to index
```
jacebrowning: The character before "Index" doesn't render for me. I just see a `[?]`.

jacebrowning:
I think the new styling looks great! Maybe the index should be wrapped in `container` to set a maximum width? https://doorstop-pr-353.herokuapp.com/index.html

michaelnt:
I think it looks good too, thanks for making the improvements.

If the content of an item, e.g. an image is very wide and you scroll the navbar overlaps the content. Do you now how to fix the navbar vertically but have it scroll off horizontally?

The navbar is wider (3) now and the main content narrower which makes the above problem worse. Even before this change we aren't using the full width for some reason. A navbar of 2 columns and a content of 10 would probably be better for my documents but perhaps other people are using longer headings? 

When I try the -b header option (I handn't noticed this before) I get blank names for items where I haven't added the header attribute yet. This seems to be the [desired behaviour](https://github.com/jacebrowning/doorstop/pull/287#pullrequestreview-100822238) but doesn't look great. I think the UID could be displayed if a header isn't defined.  I should probably open a separate Issue for that?

rickeywang:
> I think the new styling looks great! Maybe the index should be wrapped in `container` to set a maximum width? https://doorstop-pr-353.herokuapp.com/index.html

Yes I will look into that. 

> When I try the -b header option (I handn't noticed this before) I get blank names for items where I haven't added the header attribute yet.

Is this what you meant? 

Capture

I have not tested anything with very wide images, since I don't use them myself and there is no sample in the repo for these. Making the content be full width currently also makes all the text become full width. It's easier on the eyes to read multiple shorter lines of text than one long line of text, which would be why I think the template was not using full width. (just my speculation)

But for charts or other images, wider content container should be used. I think that should be a different pull request though. 

I made the navbar width 3 because it can fit it more English words comfortably, and there is more whitespace to separate nav with content. I have no strong feelings if using 2. Ideally the navbar should be collapsible/draggable such that this is a non-issue.

jacebrowning:
Additionally, we may want to pull the header concept out of "beta" so that document publishing is consistent.

jacebrowning:
@rickeywang Still working on this?

rickeywang:
I think we can close this PR in favour of #371 since @michaelnt has done some work to update the unit tests and fixed the outstanding issues.

PR #355, opened 2019-07-05
Add SPDX-License-Identifier: LGPL-3.0-only

sebhub:
https://spdx.org/licenses/LGPL-3.0-only.html


jacebrowning:
@sebhub Can you cite any reference that suggests including the license in every file is required?

sebhub:
It is not required, but it helps users of open source software to identify the open source license used in a particular source file. There are tools which automatically scan files to extract the information in case a package is embedded in a larger project.

Issue #134, opened 2014-08-14
Add user-defined "Review" artifacts


robertberrington:
Can we through a config file define other "review" items and the fields on which the hash is calculated?

I like the current "review" structure, but I'd like the flexibility to extend it to other reviews of the information for other aspects of work.

Use case ...
I'd like to define a "review artifact" called "coverage". This would have a hash calculated from the "text". I would "approve" the "coverage review" based on the following question: "Do the children-linked requirement(s) fully cover all aspects of this requirement?" Then if the requirement changes, I not only have to approve/review the requirement in it's atomic context, then review all suspect-child-links, but also answer the question of whether I need to add new requirements to "cover" the change.

There could be other use cases based on having artifacts for DER review of a DO-178 process. Maybe you can even define an artifact of artifacts (have a single artifact-hash based on a nesting of others).
##

Issue #70, opened 2014-05-22
Add a dictionary of default item values to documents.


jacebrowning:
Example `.doorstop.yml` file:

```
settings:
...
defaults:
  custom-attr1: True
  custom-attr2: "Hello, world!"
```

Example usage:

```
item = document.add_item()
assert item.get('custom-attr2') == "Hello, world!"
```
##

sebhub:
Implemented by commit 6c91fab39f9a2d5b1e46e7276e89f4e803cdde8c.

PR #359, opened 2019-07-10
Optionally add extended attr to the fingerprint

### LINKED ISSUE #337: Optionally add extended attributes to fingerprint
Currently, the [extended attributes](https://doorstop.readthedocs.io/en/latest/reference/items/#extended-attributes) are not covered by the fingerprint used by the [doorstop review](https://doorstop.readthedocs.io/en/latest/reference/items/#reviewed). It should be possible to place also extended attributes under the fingerprint scope. One way to do this could be via the `.doorstop.yml` document configuration file:
```yaml
settings:
  digits: 3
  prefix: X
  sep: ''
  reviewed: [a, b, c]
  optional: [a]
```
The `reviewed` and `optional` are new optional document configuration options.

Values in `reviewed` name extended attributes which should be placed under the fingerprint scope in this order after the standard attributes.

Values in `optional` name extended attributes which are optional, e.g. in case an extended attribute is not found in the item file no error is reported, otherwise an error is reported.

Would such a feature be acceptable?

I have some implementation remarks.

The Item.stamp() methods needs to be extended to also cover the extended attributes which are configured to be under the fingerprint scope:
https://github.com/jacebrowning/doorstop/blob/569a24eb42423708adeed4e17a19188a8e733e56/doorstop/core/item.py#L860-L866

I am not sure how I can get in an Item method to the corresponding instance of the Document class?

TingIsGitting:
Add a document configuration option (via .doorstop.yml) to specify a
list of attributes those values will be included in the item
fingerprint.

Close #337.

### DIFF
--- a/docs/reference/items.md
+++ b/docs/reference/items.md
@@ -88,9 +88,11 @@ e.g. `level: '1.10'`.
 
 ## reviewed
 
-Each item has a fingerprint. The UID of the item, the values of the
+Each item has a fingerprint. By default, the UID of the item, the values of the
 [text](items.md#text) and [ref](items.md#ref) attributes, and the UIDs of the
-[links](items.md#links) attribute contribute to the fingerprint.
+[links](items.md#links) attribute contribute to the fingerprint. Optionally,
+values of extended attributes can be added to the fingerprint through the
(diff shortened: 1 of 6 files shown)


sebhub:
I will create a new pull request that tries to address #70 and #134.

PR #366, opened 2019-07-15
Add reviews and default values to extended attributes

sebhub:


sebhub:
@jacebrowning Please let me know if you want to have it differently. This feature is very important for us. I will do whatever it takes to get it integrated. It is not a problem at all if you need more time to review it, but since you merged other patches quite quickly, I begin to wonder what is wrong with this patch set.

PR #369, opened 2019-07-18
Header

michaelnt:
Promote the header feature out of beta


michaelnt:
Sorry had some trouble getting the test environment working turns out on ubuntu at least you need to create a virtualenv first.

I left the setting so this feature could be disabled but I don't think there is is any reason why you would need to disable it, you can simply leave the header field blank.

jacebrowning:
> Sorry had some trouble getting the test environment working turns out on ubuntu at least you need to create a virtualenv first

`$ make test` should do that for you. Did  you run into any problems with that?

michaelnt:
On my desk I use windows but not git and make, and then anaconda on remote linux servers. It was painful trying to get poetry to work with the system packaged python in ubuntu (running inside WSL). If you create a virtual env first (python -m venv ./venv) and activate it then make test etc. work fine.

jacebrowning:
https://pypi.org/project/doorstop/1.6b3/

Issue #367, opened 2019-07-16
Add include feature to document configuration


sebhub:
The document configuration may define [default values for attributes](https://github.com/jacebrowning/doorstop/blob/develop/docs/reference/items.md#defaults-for-extended-attributes). To avoid excessive copy and paste if attributes are reused in different projects and documents an include feature to create the document configuration from multiple files may be helpful.

There's an example [here](https://stackoverflow.com/a/9577670/429533) of defining a custom `!include` constructor that might work for this.

A follow up feature could add initialization via custom Python code.

This feature proposal is a result of discussions in #70, #319, and #330.

sebhub:
Should the include feature work only for `.doorstop.yml` files or also for item files?

I noticed that Doorstop loads all *.yml files inside the project. There may be unrelated *.yml files present. Is it really a good idea to load them all with the unsafe yaml.FullLoader?

Issue #330, opened 2019-04-24
Custom item template


RomainTT:
Possibly embraces #70 and #319, with a little bit more.

It would be nice being able to write item templates like this:

```yaml
custom_field: custom_default_value
text: |
    Default text that appears in each new item.
```

And tell a document to use this template in `.doorstop.yml`:

```yaml
settings:
  digits: 3
  prefix: SRS
  sep: ''
  template: my_template.yml
```

Then, when I create a new item in SRS it has immediatly the content of the template:

```yaml
active: true
derived: false
level: 1.0
links: []
normative: true
ref: ''
reviewed: None
custom_field: custom_default_value
text: |
    Default text that appears in each new item.
```

Personally, I would need this to include some mandatory fields in items that are describing tests. And I want all my items to have the same fields with the same default value.

Meanwhile, I don’t know where should be located the template files, maybe anywhere since doorstop can search the whole depository?

jacebrowning:
Could the default fields be a part of `.doorstop.yml`? Maybe there could be a new `defaults` top-level dictionary?

RomainTT:
I was thinking about this too. But let’s say I want to apply a common template to many documents, because for a specific project I am writing requirements with a number of custom fields, and these requirements are located in different documents. I would be forced to copy-paste my `defaults` dictionary to every `.doorstop.yml`. That could work for a few fields, but it becomes very annoying when managing like 10 fields for instance.

TingForGitHub:
Due to ECSS standards, we need to add some more keys such as "Type", "Verification Method", etc. into the .yml files for requirements. But are custom keys tracked by the fingerprints and the review? If not, will this feature be considered to be added to a newer version of Doorstop?

We also want to have the custom keys with default value to take effect on every requirement .yml file once we define the custom keys.

jacebrowning:
Custom keys are not currently included in the fingerprint: 

https://github.com/jacebrowning/doorstop/blob/569a24eb42423708adeed4e17a19188a8e733e56/doorstop/core/item.py#L862-L864

I would be happy to review a PR that includes custom keys in the fingerprint.

sebhub:
I think this issue asks for two things:

1. A way to initialize extended attributes with document-specific default values. This is #70 

2. A way to include other YAML files in .doorstop.yml.

The second thing could be handled with something like this:
```yaml
include: absolute/or/relative/path/to/some.yml
```
Or:
```yaml
include:
- absolute/or/relative/path/to/some.yml
[...]

PR #376, opened 2019-07-19
Item templates

NouberNou:
Intent is to allow for adding new items to pull templated text into the item from template files stored in the root of the project within a directory called `.doorstop`.

This includes a default template and a new `--template` parameter to `doorstop add`.

Rationale: Too lazy to constantly copy and paste in the default template for requirement structure that my organization is using for every requirement (like subheaders for rationale, other requirement information).

Looking for recommendations on how templates should be added, the `.doorstop` directory is just something I came up with on the fly, and seemed reasonable.


NouberNou:
I am not sure what the diff failures mean in the CI tests?

michaelnt:
The make check failed.

You need to run that command, which will run the black formatter and a command to sort the imports. Then check in those format changes.

NouberNou:
Thanks!

Of course now that kicked `test_document.py` over the pylint line limit, haha!

NouberNou:
I understand module size limits in functional code, but it seems a bit overkill in test files.

NouberNou:
Updated pylint ignores for that test based on the precedence of it being done in the items test suite.

sebhub:
This looks like a duplication of the [custom default values for attributes](https://github.com/jacebrowning/doorstop/blob/develop/docs/reference/items.md#defaults-for-extended-attributes) that was added recently, see #70 and #330.

Related is a feature proposal #367.

NouberNou:
While I agree that this is a similar system at first glance, I believe that proposing something wholly owned by doorstop is the most prudent in this situation.

Do not dismiss my gratitude for such an edenvenour though as my next conquest was that of the markdown system and how YAML and MD could support a more dynamic and supportive ecosystem.

RomainTT:
Hi,
could you write some documentation about your feature please ? You want to highlight that it is different from the other issues mentioned above, so a bit of explanations with examples and so on would be nice.

I think I like your way of thinking with the concept of "template", but for now I find it a bit in conflict with the work of #366.

NouberNou:
Sure. The documentation is pretty simple:

1. Place a folder in your root project directory named `.doorstop`. This should be the same folder where you normally run the `doorstop` command from.
2. Place template files in there, nominally files with the extension `.md`, for example `my-template.md` and populate file with your custom default item text content.
3. When adding a new item with doorstop use the following argument `--template=my-template.md`

This will produce a new item with the next field populated by the contents of `my-template.md`.

Alternatively, place a file called `default.md` in the `.doorstop` folder and this will be used for any new item that does not specify the `--template` option.

NouberNou:
The intent is similar, but this is a more dynamic way of doing it. Alternatively moving to something that is more dynamic than just a single .doorstop.yml file would also be appropriate.

The problem I am addressing is that different requirements have different fields by default, and being able to create a test requirement vs. a functional requirement with its default layout has value, more so than one single default value.

NouberNou:
Any more thoughts on this?

RomainTT:
I don’t know what the others think, but in my opinion there should not be both implementations from 6c91fab39f9a2d5b1e46e7276e89f4e803cdde8c **and** this PR.

I personally prefer the way of this PR concerning the management of defaults values. Therefore, my suggestion is to revert 6c91fab39f9a2d5b1e46e7276e89f4e803cdde8c in order to replace it by this new implementation.

Note that the documentation for this PR should be present in `docs/reference/items.md`, and would replace the one from 6c91fab39f9a2d5b1e46e7276e89f4e803cdde8c.

Also note that this replacement would still be compatible with the feature from c8834c87003e32b64748d00e296243c077df5c0b (reviewed extended attributes).

I don’t know if an actual `git revert` is enough, as many commits were pushed since.

sebhub:
The existing support for default attributes is document-specific. I would not remove this feature. I think this commit series should be refactored to better use the infrastructure that is already in place. You can for example apply the command-specific default attributes after the document-specific default attributes with a second call to Item.set_attributes(). For the project-specific default attributes, I would add them to the top level `.doorstop.yml` or solve it via the include feature #367 in favour of adding a new `.doorstop` file.

sebhub:
With the include support you could simply place this into the `.doorstop.yml`:
```yaml
settings:
  ...
attributes: !include template.yml
```
In `template.yml`:
```yaml
text: 'Your template text'
```

sebhub:
You can find a proof of concept for the include feature here:

https://github.com/sebhub/doorstop/commit/90482341622554d959bf9c0715379f8005315c75

sebhub:
Please check if #403 does what you need.

NouberNou:
Sorry, was away over the weekend. I'll try to get a look at it today.

NouberNou:
Yea that looks good, but I'd highly recommend describing in the documentation what a `default.yml` file should look like. Does it have to be an exact copy of a requirements yml? Can it specify only the fields wanted and it'll just override those?

How does it differ from the defaults in the .doorstop file?

It looks like it is functionally the same, but less clear on its usage, which has a been a problem in general with Doorstop documentation that I am getting reported by a number of people in my organization that are evaluating it right now.

I'll close this PR for now.

PR #382, opened 2019-07-24
Document scripting interface

### LINKED ISSUE #381: Document the scripting interface
Include same code from https://speakerdeck.com/jacebrowning/doorstop-requirements-management-using-python-and-version-control?slide=37.

jacebrowning:
Closes #381

### DIFF
--- a/docs/api/scripting.md
+++ b/docs/api/scripting.md
@@ -1,1 +1,75 @@
-> Help wanted! Please [contribute to this documentation](https://github.com/jacebrowning/doorstop/edit/master/docs/api/scripting.md).
+<h1>Scripting Interface</h1>
+
+Being written in Python, Doorstop allows you to leverage the full power of Python to write scripts to manipulate requirements, run custom queries across all documents, and even inject your own validation rules.
+
+# REPL
+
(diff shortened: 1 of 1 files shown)


docs/api/scripting.md
```diff
@@ -1 +1,75 @@
...
+#!/usr/bin/env python
+
+
+import sys
+from doorstop import build, DoorstopInfo, DoorstopWarning, DoorstopError
+
+
+def main():
+    tree = build()
+    success = tree.validate(document_hook=check_document, item_hook=check_item)
```
jacebrowning: I'll open an issue after this is merged, but the idea of a "plugin" system is that these callbacks could be passed automatically if a certain convention is met (e.g. `plugin.py` exists which defines `check_document` or `check_item`).

sebhub: For a plugin system it would be nice to be able to register custom sub-commands to doorstop with custom sub-command arguments.

I think you need also a generic tree iterator which doesn't perform the default validations.

jacebrowning: Yeah, I could see that being useful. I'm going to merge this as-is, which documents the existing functionality.

PR #384, opened 2019-07-25
Make the item stamp algorithm configurable

sebhub:


jacebrowning:
Thanks for looking into using different algorithms, but I don't want the algorithm to be configurable. Users should not be burdened with a decision that is really only an implementation detail. If the concern is that `md5` might result in collisions then we should simply use a different algorithm in Doorstop 2.0.

sebhub:
Ok, what about the first two patches? Should I create a new pull request for them?

Do you have a schedule for the 2.0 release? The cryptographically insecure MD5 has is a blocking point at the moment for me. I abandoned the approach with the digital signatures.

jacebrowning:
> Should I create a new pull request for them?

Yes, please do.

> Do you have a schedule for the 2.0 release? 

I don't have a specific date in mind, but it can come shortly after the 1.6 release: https://github.com/jacebrowning/doorstop/milestone/14

sebhub:
Ok, I will have a look at the bugs, but I am not really an HTML/CSS expert.

PR #388, opened 2019-07-28
doorstop/gui: extract creation of properties to a separate widget function

stanislaw:
Encapsulating the details of the 'properties' widget in a single function
will help to manipulate this widget easier in the future.

If this kind of change is welcomed, I will create a similar changeset for the widget with the details about viewing, creating and removing links. When that is also done, I can start introducing a widget for the multiple refs case.

### DIFF
--- a/doorstop/gui/main.py
+++ b/doorstop/gui/main.py
@@ -435,21 +435,9 @@ class Application(ttk.Frame):
             self.text_item.grid(row=1, column=0, columnspan=3, **kw_gsp)
 
             # Column: Properties
-            widget.Label(frame, text="Properties:").grid(
-                row=2, column=0, sticky=tk.W, **kw_gp
+            self.create_properties_widget(frame).grid(
+                row=2, rowspan=5, column=0, columnspan=1, sticky=tk.NSEW, **kw_gp
             )
-            widget.Checkbutton(frame, text="Active", variable=self.intvar_active).grid(
-                row=3, column=0, sticky=tk.W, **kw_gp
-            )
-            widget.Checkbutton(
-                frame, text="Derived", variable=self.intvar_derived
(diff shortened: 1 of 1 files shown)


stanislaw:
@jacebrowning @sebhub @elarivie

I have just noticed this PR and scanned through it with great interest: https://github.com/jacebrowning/doorstop/pull/301.

With all due respect to the work done by @elarivie, it looks like the changeset suggested there is rather massive and the fact, that it has been opened one year ago and has not been merged since then, indicates that it could be more practical to:

1) find a good venue to discuss a general concept for the GUI tool and agree on the common set of features to make the existing GUI more useable.

2) create a roadmap and tickets for the changes that are missing in the @elarivie's changes and split those changes that are present there into much smaller pull requests.

From my side, this pull request #388 I have just opened is based on the idea of doing very small incremental changes for any upcoming GUI work. What I observe is that as a step zero, the UI layout's modularity could be improved and this would simplify both local reasoning about the UI elements and high-level picture of how the layout works.

I have many ideas on how to restructure and improve existing layout but I think this has to be a coordinated effort spread over a sequence of very small GUI improvements. Each of the changes has to be discussed, reviewed and aligned with a high-level vision in mind: which direction the Doorstop GUI has to advance to.

jacebrowning:
> If this kind of change is welcomed, I will create a similar changeset for the widget with the details about viewing, creating and removing links. 

Extracting code like this into functions is much appreciated!

> I have just opened is based on the idea of doing very small incremental changes for any upcoming GUI work.

I like this approach.

> Each of the changes has to be discussed, reviewed and aligned with a high-level vision in mind

Perhaps someone would like to open multiple feature requests that are screenshots of the current GUI mocked up with the proposed change?

PR #392, opened 2019-07-29
doorstop/gui: extract creation of ref-related elements to a separate widget function

stanislaw:
One more small intermediate PR, the next one will be the "many refs" one.

### DIFF
--- a/doorstop/gui/main.py
+++ b/doorstop/gui/main.py
@@ -443,12 +443,10 @@ class Application(ttk.Frame):
             )
 
             # External Reference
-            widget.Label(frame, text="External Reference:").grid(
-                row=7, column=0, columnspan=3, sticky=tk.W, **kw_gp
+            self.create_reference_widget(frame).grid(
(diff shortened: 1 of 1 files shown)


doorstop/gui/main.py
```diff
@@ -443,12 +443,10 @@ def text_extendedvalue_focusout(event):
             )
 
             # External Reference
-            widget.Label(frame, text="External Reference:").grid(
-                row=7, column=0, columnspan=3, sticky=tk.W, **kw_gp
-            )
-            widget.Entry(frame, width=width_text, textvariable=self.stringvar_ref).grid(
-                row=8, column=0, columnspan=3, **kw_gsp
+            self.create_reference_widget(frame).grid(
+                row=7, rowspan=2, column=0, columnspan=2, sticky=tk.NSEW, **kw_gp
```
stanislaw: it is not a regression: it does not need to be a `columnspan=3` because here only two columns are configured:

https://github.com/jacebrowning/doorstop/pull/392/files#diff-b07e5654f6b3e8f988965d73baa74cefL401

PR #395, opened 2019-08-01
WIP: RFC: Item#ref: initial support of many ref items

stanislaw:
This work is ready to be reviewed. Please also review the UI changes.

From my side, I am mainly concerned with the following thing:

I have followed the plan suggested by @sebhub [here](https://github.com/jacebrowning/doorstop/issues/365#issuecomment-511099791) of extending the `ref` to support both original string behavior and new array behavior, in order to preserve the backward compatibility. The outcome of this is that the code related to handling of the ref attribute is now always branching: 

```
if array do sth 
else if string do sth
# or
if str do sth
else if array do sth
```

Also, you can see this in the updated docs: having `ref` as both string and array maybe confusing.

After looking at this final work, I am thinking of whether we should rather create another attribute with a plural name like `references` and move new behavior there. Then it will be easier to still support the original behavior but also migrate to the new behavior by switching `ref -> references`. If `ref` is used, the system works in the original way, if `references` is used the system works in the new way. If both are specified, the new behavior is preferred.

IMO, the switch from using one field to another would be an easier way for users to migrate and also a cleaner implementation.

I am looking forward to your feedback.

Example: 

```yaml
references:
  - path: external/text.txt
     type: file
  - path: external/text2.txt
     type: file
```

P.S. The UI has no tests so it is rather hard to iterate on it and have a good confidence that everything works :(


jacebrowning:
I didn't mention this in the issue, but I have also thought that perhaps we should use a new name for "multiple references". 😄 

I'd say go ahead and make that change as you have proposed above. This will also make the new logic easier to review as we can be sure the existing behavior remains unchanged.

stanislaw:
Alright, I will change it then. It should go rather fast since I know all of the pivotal points now. Also, it will be possible to preserve the UI behavior by supporting the old widget when it is a string and the new ListBox one when it's `references`.

sebhub:
I would maintain all variants in one attribute (`ref`). To get rid of the ifs you could use a new class Reference and use virtual methods instead of ifs. This would also concentrate the implementation of different variants in one place and not throughout the code.

jacebrowning:
We can also consider dropping `ref` altogether in a future major release to only support the new format.

sebhub:
Ok, I still would prefer [polymorphism over conditionals](https://refactoring.guru/replace-conditional-with-polymorphism), but this can be also done in a follow up patch.

stanislaw:
@jacebrowning it is much easier to make changes with the separate `references` field, however the question arises when it comes to updating the export/import tests:

For example, should I add both columns `ref` and `references` to the CSV/TSV/XLS at the same time even if only one of these fields is set in a given item? Is it ok to keep these two fields for a while, until the `ref string` behavior is deprecated?

---

@sebhub,

> I would maintain all variants in one attribute (ref). To get rid of the ifs you could use a new class Reference and use virtual methods instead of ifs. This would also concentrate the implementation of different variants in one place and not throughout the code.

I now see the pros and cons for both approaches but am rather voting for the separate `references` PR. It is not only about branching ifs everywhere, but also consistent user experience. Adjusting the GUI to handle both `ref string` and `ref array` at the same time is a bit tricky:

at which point the CLI and GUI shall decide that new `ref: [array]` shall be written to the YAML that had `ref: string` before? 

We agreed that `ref array` supports files only for now (type: file) so the behavior is currently quite different from the existing behavior of the `ref string` case.

Another example is `find_ref` method: according to the existing behavior it returns `(path,  line or None)` tuple, while the new implementation an array of paths (and in the future even more complex details). I think it is better to do the refactorings outside the scope of this change, so simply adding the second field and adding the new strategy such as `find_references` would be easier.

Please check this PR's diff. I think I have handled all of these cases decently, so maybe given that @jacebrowning considers deprecating the `one ref` behavior, the implementation of this PR could be an a good intermediate step before deprecating that behavior. Again, I see pros and cons in both approaches.

michaelnt:
Could the old field be upgraded to the new behaviour automatically when you run the new version of doorstop. Or refuse to run until an upgrade command is run.

Could add a version field to make this easier in the future.

stanislaw:
> Could the old field be upgraded to the new behavior automatically when you run the new version of doorstop. Or refuse to run until an upgrade command is run.

Within the scope that we discussed, this is not possible easily: the `ref` has a functionality of finding a given keyword in some file, while the new implementation of `references` links to the files directly. There was a comment about also supporting keywords with the new `references` but afaik the behavior will also be slightly different.

stanislaw:
> @jacebrowning it is much easier to make changes with the separate references field, however the question arises when it comes to updating the export/import tests:

> For example, should I add both columns ref and references to the CSV/TSV/XLS at the same time even if only one of these fields is set in a given item? Is it ok to keep these two fields for a while, until the ref string behavior is deprecated?

@jacebrowning these are really the last 7 tests that I haven't fixed yet: CSV, TSV, XLS. For HTML/Markdown which are not tabular, this is not a problem.

stanislaw:
One possible policy to resolve having both `ref` and `references` columns in CSV, TSV, XLS would be:

When exporting multiple items to a document:

- 1) If there is no a single reference in `references` in all items, only the `ref` column is exported.
- 2) If there is no a single reference in `ref` in all items, only the `refeferences` column is exported.
- 3) Otherwise both columns are exported.

Point 1) will make it easier for existing requirements based on ref to not be forced to make any changes
Point 2) will make it clean for the users of the new implementation because they are not going to use old `ref` behavior.
Point 3) can make it clear where the migration is still needed.

What do you think?

jacebrowning:
@stanislaw Yeah, that sounds good to me. If it reduces possible conflicts, you can make it an error to attempt to use both `ref` and `references`.

PR #397, opened 2019-08-02
Use Text.save_text() for all string attributes

sebhub:
The default YAML library formatting for strings is quite ugly.  Use the
nice Text.save_text() formatting for all string valued attributes.

### DIFF
--- a/doorstop/core/item.py
+++ b/doorstop/core/item.py
@@ -25,6 +25,32 @@ from doorstop.core.types import UID, Level, Prefix, Stamp, Text, to_bool
 log = common.logger(__name__)
 
 
+def _convert_to_yaml(indent, prefix, value):
+    """Convert value to YAML output format.
+
+    :param indent: the indentation level
+    :param prefix: the length of the prefix before the value, e.g. '- ' for
+    lists or 'key: ' for keys
+    :param value: the value to convert
+
+    :return: the value converted to YAML output format
+
+    """
+    if isinstance(value, str):
+        length = indent + prefix + len(value)
+        if length > settings.MAX_LINE_LENGTH or '\n' in value:
+            value = Text.save_text(value.strip())
+        else:
+            value = str(value)  # line is short enough as a string
+    elif isinstance(value, list):
+        value = [_convert_to_yaml(indent, 2, v) for v in value]
+    elif isinstance(value, dict):
+        value = {
+            k: _convert_to_yaml(indent + 2, len(k) + 2, v) for k, v in value.items()
+        }
+    return value
+
+
 def requires_tree(func):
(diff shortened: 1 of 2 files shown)


sebhub:
Some background information why this change is useful. I would like to write interface specifications with Doorstop and use the data to generate header files with Doxygen markup, content for a user manual, and content for an Interface Control Document mandated by standard. The Doorstop item could look like this:
```yaml
active: true
derived: false
header: ''
level: 1.1
links: []
normative: true
ref: ''
reviewed: db5392776451fbfba307a291300190f0
order:
- RTEMS-Classic-Event-Receive-001
type: application-interface
text: N/A
interface:
  function:
    return:
      type-link: RTEMS-Classic-Types-001
      values:
      - link: RTEMS-Classic-Event-Receive-002
        value-link: RTEMS-Classic-Types-004
        description: event set sent successfully
      - link: RTEMS-Classic-Event-Receive-003
        value-link: RTEMS-Classic-Types-005
        description: invalid task id
    designator: rtems_event_send
    parameters:
    - type-link: RTEMS-Classic-Types-002
      description: |
        Identifier of the target task. Specifying RTEMS_SELF results in the
        event set being sent to the calling task.
      designator: id
      order: 0
    - type-link: RTEMS-Classic-Types-003
      description: |
        Event set sent to the target task.
      designator: event_in
      order: 1
    description: |
      This directive sends an event set, event_in, to the task specified by id.
      If a blocked task’s input event condition is satisfied by this directive,
      then it will be made ready. If its input event condition is not
      satisfied, then the events satisfied are updated and the events not
      satisfied are left pending. If the task specified by id is not blocked
      waiting for events, then the events sent are left pending.
    notes: |
      Specifying RTEMS_SELF for id results in the event set being sent to the
      calling task.

      Identical events sent to a task are not queued. In other words, the
      second, and subsequent, posting of an event to a task before it can
      perform an rtems_event_receive has no effect.

      The calling task will be preempted if it has preemption enabled and a
      higher priority task is unblocked as the result of this directive.

      Sending an event set to a global task which does not reside on the local
      node will generate a request telling the remote node to send the event
      set to the appropriate task.
```
The problem is that if I hand over this file to `doorstop review`, then the nested text attributes are quite ugly formatted:
```yaml
interface:
  function:
    description: 'This directive sends an event set, event_in, to the task specified
      by id.

      If a blocked task’s input event condition is satisfied by this directive,

      then it will be made ready. If its input event condition is not

      satisfied, then the events satisfied are updated and the events not

      satisfied are left pending. If the task specified by id is not blocked

      waiting for events, then the events sent are left pending.

      '
    designator: rtems_event_send
```

PR #399, opened 2019-08-05
RFC: Item#references: initial support of many ref items (Take 2)

stanislaw:


doorstop/core/item.py
```diff
@@ -163,6 +164,14 @@ def _set_attributes(self, attributes):
                 value = Text(value)
             elif key == 'ref':
                 value = value.strip()
+            elif key == 'references':
+                if not isinstance(value, list):
```
sebhub: I would not silently accept this. I would throw an exception.

docs/reference/item.md
```diff
@@ -177,7 +177,43 @@ item of the reviewed attribute, not the parent item of the link).  The
...
-## `ref`
+## `references` (new array behavior)
+
+An array of external references. An item may reference a number of external
+files. The external references are displayed in a published document.
+
+Doorstop will search in the project root for a file matching the specified
+reference. If multiple matching files exist, the first found will be used.
+
+A file is considered a text-file unless its file extension is listed in
```
sebhub: For what is this text-file stuff necessary?

sebhub:
Looks good, with all this import/export and GUI stuff it was quite a bit of work.

doorstop/core/item.py
stanislaw: I expected this kind of a comment to ask a more general question: what is the best place to validate the `references` parameter?

In addition to checking that the input is a list, more things could be checked, such as:

- each element of the `references` list shall be a dictionary
- `type` has to be of a `file` value (currently the one supported)
- `path` shall be a non-empty string

Do we want to include all of these here this MR or a simple exception like @sebhub is suggesting is enough?

docs/reference/item.md
stanislaw: I don't think we need this for the `references` feature, even though I think most of the time we will link text files which are easier to check than binaries.

No strong opinion about this.

doorstop/core/item.py
sebhub: I think the program should reject all unexpected input. If you silently ignore this (e.g. because a future version of Doorstop added new things which you don't understand), then this may cause unexpected effects which are hard to trace for a user.

I would iterate through the list and empty the dictionaries with pop(). If exceptions occur, then print a message and terminate.

docs/reference/item.md
sebhub: I think in this commit, we don't look at the content?

stanislaw:
@sebhub, I have incorporated both of your comments and now I am again hitting the 1000 limit but this time it is `item.py` file with a rather huge Item class. I have around 1030 lines by now.

Any advice on the quick refactoring of the Item class in a separate PR so that I could proceed here?

I think the details of YAML reading/writing could be extracted to a separate file: `yaml.py` with YamlWriter and YamlReader classes.

Or, the validation could be extracted to a separate class like ItemValidator. Here I mean the functionality of the methods `get_issues` and the methods it uses.

What would you suggest?

@jacebrowning @sebhub

sebhub:
I would add a Reference base class (maybe to a separate file) and add specializations, e.g. KeywordReference (the old behaviour), FileReference, DirectoryReference, etc. Upon item load, specializations are constructed based on the attribute values. The Reference base class should have virtual methods for import, export, validation and publish.

stanislaw:
@sebhub I would suggest that introducing more class hierarchies should be done as a separate work outside of the scope of this PR.

For the scope of just freeing spare 30 lines, I thought about some easier refactoring that only moves some functionality to a separate file such as YAML reading/writing or validating.

Maybe there is something even easier.

stanislaw:
@jacebrowning @sebhub 

Please check my latest updates. I have relaxed the size of the Item document by introducing two classes:

1) ReferenceFinder (helper class responsible for finding references)
2) ItemValidator (class responsible for validation of the Item's scheme). Now it is obvious the other fields can be checked in a similar way too.

Let me know if this makes sense to you.

P.S. @sebhub I have also checked the feasibility of doing the change: `KeywordReference (the old behaviour), FileReference, DirectoryReference, etc` and I think doing that change would not help me to overcome the Item's huge size problem.

docs/reference/item.md
```diff
@@ -177,7 +177,40 @@ item of the reviewed attribute, not the parent item of the link).  The
...
+```
+
+### Note: new behavior vs old behavior
+
+Note: Newer `references` attribute is a new array behavior compared to the 
+original `ref` attribute's string behavior. The old behavior supports referencing
+only one file via file name or referencing a file that contains a given
+keyword.
+
+The new behavior of `references` attribute allows referencing many files. It
```
stanislaw: @jacebrowning @sebhub 

Is it ok to keep the scope of this PR as it is without implementing the `keyword` to also support the old `ref` behavior?

I am fine with both implementing it here or doing a separate PR. In both cases, the Item has to be refactored because it is now again close to being 1000 lines. 

Having the above said, I think the PR is ready for review in its current shape. Please take a look.

PR #407, opened 2019-08-11
Switch to SHA1 for detecting changes to files

### LINKED ISSUE #358: MD5 as a stamp algo
I'm new to doorstop and find it very interesting. I'm a bit concerned about the fact that fingerprinting as per Stamp class implementation here:

https://github.com/jacebrowning/doorstop/blob/5f624c80f81c950fedc30af34852979b67d57597/doorstop/core/types.py#L508) 

is still using MD5 hashing algorithm. Apart from being deprecated since long ago for any cryptography use (which is not the case here) it's also something not easily found on any modern application. Git itself is using SHA1 for creating hashes, and I suspect any expert body reviewing doorstop would regard using MD5 as suspicious as a minimum. Collision is not far from happening, and mangling for causing collision is also possible. Before the develpment goes too far would it be a good idea to (1) switch to SHA1 and (2) include a reference to the used algorithm in each hash so that any future change would not break compatibility?

jacebrowning:
Resolves #358 

Note that this is a breaking change so it will require a bump in the major version number to be released. I don't think we need any higher level of hash collision prevention that what Git uses.

cc @stefanoco


sebhub:
SHA1 is also an insecure hash algorithm. I would use SHA256.

jacebrowning:
@sebhub This feature is not tied to security and I don't want anyone to get the idea that the file fingerprinting is any form of cryptographic protection against changes. I'd like to keep these fingerprint values as short as possible so they fit into narrow editor and terminal windows. SHA1 is sufficient for tracking content in Git, so why shouldn't the same be true here?

stefanoco:
Actually when commenting on SHA1 vs MD5 my main concern was trying to align with some already working solution to a similar problem (tracking files changes for functional reasons, not for security) being MD5 relatively week in this context, thus SHA1 as per git implementation. Maybe this in the future will enable some interesting integration.

sebhub:
> @sebhub This feature is not tied to security and I don't want anyone to get the idea that the file fingerprinting is any form of cryptographic protection against changes. I'd like to keep these fingerprint values as short as possible so they fit into narrow editor and terminal windows. SHA1 is sufficient for tracking content in Git, so why shouldn't the same be true here? smile

There are already plans to introduce a new hash function in Git:

https://github.com/git/git/blob/master/Documentation/technical/hash-function-transition.txt

In 2018, SHA256 was selected as the new hash function:

https://github.com/git/git/commit/0ed8d8da374f648764758f13038ca93af87ab800#diff-362a5bda8e8b2aac7bd08f6948550f89

Replacing one insecure hash with another makes little sense. With respect to random collisions the MD5 is still perfectly fine. If you are concerned about the length of the digest, then you can also use a base64 encoding:
```patch
diff --git a/doorstop/core/types.py b/doorstop/core/types.py
index cfcd36e..d87b269 100644
--- a/doorstop/core/types.py
+++ b/doorstop/core/types.py
@@ -2,6 +2,7 @@
 
 """Common classes and functions for the `doorstop.core` package."""
 
+import base64
 import hashlib
 import os
 import re
@@ -560,10 +561,10 @@ class Stamp:
     @staticmethod
     def digest(*values):
         """Hash the values for later comparison."""
-        md5 = hashlib.md5()
+        hsh = hashlib.sha256()
         for value in values:
-            md5.update(str(value).encode())
-        return md5.hexdigest()
+            hsh.update(str(value).encode())
+        return base64.b64encode(hsh.digest()).decode('utf-8')
 
 
 def to_bool(obj):
```
This leads to changes such as this:
```patch
diff --git a/doorstop/cli/tests/docs/HLT001.yml b/doorstop/cli/tests/docs/HLT001.yml
index d155697..182285b 100644
--- a/doorstop/cli/tests/docs/HLT001.yml
+++ b/doorstop/cli/tests/docs/HLT001.yml
@@ -11,6 +11,6 @@ links:
 - TUT019: 85c96dfa447540b5433a67f1e1114c61
 normative: true
 ref: test_tutorial_section_1
-reviewed: 2e4702d52e9ab2fd24b9d22270bcc4dc
+reviewed: LuZMejVVbNoVI+fZff7mrEqiORKLfSvW0IYGX1CTafA=
 text: |
   Tutorial Section 1.0:
```

jacebrowning:
Moved to https://github.com/jacebrowning/doorstop/pull/410

PR #408, opened 2019-08-11
Enable mypy for checking type annotations

jacebrowning:
I've disabled the type checker for GUI files as I consider those critical at the moment. Anyone working on those files is welcome to attempt to add proper type annotations.


doorstop/cli/main.py
```diff
@@ -74,7 +74,9 @@ def main(args=None):  # pylint: disable=R0915
     }
 
     # Build main parser
-    parser = argparse.ArgumentParser(prog=CLI, description=DESCRIPTION, **shared)
+    parser = argparse.ArgumentParser(  # type: ignore
+        prog=CLI, description=DESCRIPTION, **shared
+    )
```
jacebrowning: I am not sure why `mypy` emits a warning for this, but this code is clearly working, so I've disabled it.

doorstop/common.py
```diff
@@ -22,12 +22,11 @@ def _trace(self, message, *args, **kws):
 
 
 logging.addLevelName(logging.DEBUG - 1, "TRACE")
-logging.Logger.trace = _trace
+logging.Logger.trace = _trace  # type: ignore
```
jacebrowning: We might consider converting all "trace" logging to debug.

PR #410, opened 2019-08-12
Switch to SHA256 for detecting changes to files

### LINKED ISSUE #358: MD5 as a stamp algo

jacebrowning:
Resolves #358 

Note that this is a breaking change so it will require a bump in the major version number to be released.


doorstop/core/types.py
```diff
@@ -558,12 +559,12 @@ def yaml(self):
...
-    def digest(*values):
+    def digest(*values) -> str:
         """Hash the values for later comparison."""
-        md5 = hashlib.md5()
+        sha = hashlib.sha256()
         for value in values:
-            md5.update(str(value).encode())
-        return md5.hexdigest()
+            sha.update(str(value).encode())
+        return b85encode(sha.digest()).decode('utf-8')
```
jacebrowning: Base85 is more concise, but still safely storable in YAML files.

sebhub:
Thanks for changing the hash to SHA256.

I would use base64 encoding instead of base85 since this makes it a bit easier/safer for other consumers to use the value as is, e.g. in reports. It is just four bytes more compared to base85. My choice would be urlsafe_b64encode(). You can directly grep this value for example.

sebhub:
Is it really safe to use base85 with YAML. I think something like this
```yaml
reviewed: !!foobar...
```
will be interpreted as a tag leading to a "could not determine a constructor for the tag" YAML parser error.

sebhub:
I would like to start using doorstop for some real work and the secure hash is important for me. I would like to add a new pull request which

1. changes the hash algorithm to SHA256, and

2. uses URL-safe base64 encoding.

jacebrowning:
@sebhub Sounds good! I'll close this one.

Issue #409, opened 2019-08-11
Item class needs refactoring


stanislaw:
## Motivation

- While working on #399, I have been hitting the 1000 line limit for item.py file. I have extracted some small parts to separate files but I foresee that any other addition to this file will hardly be possible.
- While working on #399, I have been rebasing with the updated `develop` a number of times and noticed that conflicts are often found in the test files related to `item` which is another indicator that the Item does too much so it creates contention easily.

## Item class overview

I have done a quick inspection of what the contents of the current `item.py` are. Here is the mostly complete overview:

- class `Item`
  - is a model-type class that has its internal exposed as properties. Examples: `item.ref`, `item.level`, `item.uid` etc.
  - (**!**) has the functionality of reading from YAML and writing to YAML (`load` and `save`)
  - (**!**) can find its parents by accessing `tree` object (`parent_items`)
  - (**!**) `Open the item for editing` functionality (`edit`)
  - linking/unlinking parent items
  - (**!**) `get_issues` validation method, checks item itself, accesses `self.tree` and `self.document` to get external information to check links (rather big methods `_get_issues_document` and `_get_issues_tree` and `_get_issues_both`)
  - `find_ref` and new `find_references` methods that find files references by `ref` and `references` fields
  - `find_child_links`, `find_child_items`, `find_child_documents`, `_find_child_objects`
  - `stamp` method
- class `UnknownItem`

I think, that the items marked with (**!**) are the most obvious candidates for extraction. Given that this project follows this very useful limitation of 1000 lines to not let files/classes grow into God Objects, it is worth to let some other classes to do jobs like validation of the item's tree or creating an item from a YAML file. This will also reduce the size of the corresponding test files.

If an agreement is found on how to proceed, one could start sending focused refactoring PRs on each agreed item.

sebhub:
I think all the validation aspects could move to a separate module. There is already some support for visiting documents and items through the hook function which I used for the cycle detection #347. A general purpose support to visit documents and items could be also handy for the plugin interface #386.

neerdoc:
@jacebrowning This could be closed and instead add a new issue that handles a more general re-factoring since there is a quite a lot already identified by the _scrutinizer_ checks.

jacebrowning:
Yeah, I agree. https://github.com/doorstop-dev/doorstop/pull/571 helped a good bit and more refactoring PRs like that would improve the codebase.

PR #419, opened 2019-09-07
ItemValidator class to encapsulate validations of Item

stanislaw:
The Item class is very close to hitting 1000 lines limit which indicates that it probably does too many things.

As discussed in https://github.com/doorstop-dev/doorstop/issues/409, which identified some possible targets for refactoring, the validation aspects could be moved to a separate module.

This changeset moves all 4 validation-related functions to a separate class: `ItemValidator`.

With this change, `Item` is no longer `BaseValidatable` which seems reasonable, because otherwise it would be hard to move these 4 functions out without introducing cyclic dependencies.

### DIFF
--- a/doorstop/core/document.py
+++ b/doorstop/core/document.py
@@ -23,6 +23,7 @@ from doorstop.core.base import (
 )
 from doorstop.core.item import Item
 from doorstop.core.types import UID, Level, Prefix
+from doorstop.core.validators.item_validator import ItemValidator
 
 log = common.logger(__name__)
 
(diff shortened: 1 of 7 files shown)


stanislaw:
It took a few approaches to the code to find the right seam to cut. Eventually, discarding the previous solution based on `BaseValidatable` was the only way to move forward. I foresee that all other classes eventually have to follow the same path and `BaseValidatable` will no longer exist.

If this change is accepted, a similar change can be done for all of the `find*` functions that could be moved to some `ItemFinder` class.

PR #423, opened 2019-10-04
Item#references: initial support of many ref items (Take 3)

stanislaw:
I am opening this as a separate MR again because the previous one received too many comments.

I really need some collaboration from the maintainers because the scope is getting quite big. Unfortunately, I don't see anything meaningful that I could split out to become separate MRs. 

Here are major comments:

1) This MR has optional `keyword` implemented as was suggested by @sebhub 👍 

> The keyword could be added to file type references in addition to the lines and stamp support. If keyword is present, then the keyword must be contained in the file content.

2) I don't know how to fit the full `references` into the Python GUI given that we also want to track the optional `keyword` there.  How many people actually use the Python GUI and how much should we care about 100% match between the command-line and GUI versions?

3) Do we really need a `type: file` part or just having `path` is enough? Can we imagine anything else except references to single files? I am quite convinced that referencing `folder`s is not practical and should always be replaced by referencing single files.

```
 references:
- path: tests/test1.cpp
  type: file
  keyword: REQ1
- path: tests/test2.cpp
  type: file
  keyword: REQ2
```

4) I am seriously suspicious about my Python writing 😅  I am happy to learn from you about every good practice that I might be missing.


sebhub:
> I am opening this as a separate MR again because the previous one received too many comments.
> 
> I really need some collaboration from the maintainers because the scope is getting quite big. Unfortunately, I don't see anything meaningful that I could split out to become separate MRs.
> 
> Here are major comments:
> 
>     1. This MR has optional `keyword` implemented as was suggested by @sebhub +1
> 
> 
> > The keyword could be added to file type references in addition to the lines and stamp support. If keyword is present, then the keyword must be contained in the file content.

I would generalize this to `regex`.

[...]
>     2. Do we really need a `type: file` part or just having `path` is enough? Can we imagine anything else except references to single files? I am quite convinced that referencing `folder`s is not practical and should always be replaced by referencing single files.
[...]
I would keep the `type` key, since this easily allows future extensions. You can use a map of constructors, e,g.
```
ctors = { 'file': FileReference.new, 'directory': DirectoryReference.new }
file_ref = ctors[type](data)
```
A directory reference could be useful if you want to ensure that a directory is completely unmodified, e.g. also no new files.

stanislaw:
> >     1. This MR has optional `keyword` implemented as was suggested by @sebhub +1

> >        The keyword could be added to file type references in addition to the lines and stamp support. If keyword is present, then the keyword must be contained in the file content.

> I would generalize this to regex.

Please see my change [here](https://github.com/doorstop-dev/doorstop/pull/423/commits/849e188714f6eb30d876d0aab6c4fac7a6727fb1). Does it work like this?

doorstop/core/reference_finder.py
```diff
@@ -89,7 +89,7 @@ def find_file_reference(ref_path, root, tree, item_path, keyword=None):
                     continue
 
                 log.debug("searching for ref '{}'...".format(keyword))
-                pattern = r"(\b|\W){}(\b|\W)".format(re.escape(keyword))
+                pattern = r"(\b|\W){}(\b|\W)".format(keyword)
```
sebhub: After looking at this piece of code I think we should keep the `keyword` attribute as is. We should add a `regex` attribute which is used as is as a regular expression without these automatic prefix and postfix additions.

stanislaw: This sounds good to me. I also was not sure about mixing the existing behavior.

stanislaw:
> After looking at this piece of code I think we should keep the keyword attribute as is. 

I have reverted the keyword regex change by simply removing that commit.

>  I would keep the type key, since this easily allows future extensions. You can use a map of constructors, e,g.

As of your other change I agree that we should keep the `type` field to accommodate for the future `directory` type feature but that should be a separate work because the scope of this MR is quite big already.

Having this said, I suggest that you guys @sebhub @jacebrowning really give it some time of code review and manual testing to see if this makes sense because I think we are very close to merging it.

One of the follow-up MRs could be to introduce a deprecation notice for people who might be using the `ref` feature.

jacebrowning:
@stanislaw I don't have much bandwidth to review changes, but if this responses to all of @sebhub's feedback them I'm good with moving forward on this.

stanislaw:
Wow! That's very good news. Thanks @sebhub.

@jacebrowning I assume we are good to merge?

jacebrowning:
@stanislaw Yeah, feel free to merge approved PRs.

PR #425, opened 2019-10-04
Fix item reformat during validation

sebhub:
The ItemValidator._get_issues_tree() did write completely wrong links
back to the item, since the item variable was re-used as a loop
variable (now: parent).  This bug did not manifest badly since
Item.set_links() had no auto_save decorator.  Remove the superfluous
Item.set_links() and use the setter instead.

### DIFF
--- a/doorstop/core/item.py
+++ b/doorstop/core/item.py
@@ -552,9 +552,6 @@ class Item(BaseFileObject):  # pylint: disable=R0902
     def is_reviewed(self):
         return self._data['reviewed']
 
-    def set_links(self, links):
-        self._data['links'] = links
-
     @requires_tree
     def find_ref(self):
         """Get the external file reference and line number.
--- a/doorstop/core/tests/test_item_validator.py
+++ b/doorstop/core/tests/test_item_validator.py
@@ -104,6 +104,45 @@ class TestItemValidator(unittest.TestCase):
         self.assertTrue(self.item_validator.validate(self.item))
         self.assertEqual('abc123', self.item.links[0].stamp)
 
+    @patch('doorstop.settings.REFORMAT', True)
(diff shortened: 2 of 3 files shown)


stanislaw:
@sebhub thanks for catching this. This behavior of 1) finding and issue and also 2) reformatting the item in the same method was something I wanted to raise later and refactor as a violation of SRP but I also managed to introduce a bug there 😞 

I want to provide a test that ensures your change is included in the test harness. I will do it today or tomorrow max.

sebhub:
I think the drop in code coverage due to the patch is harmless. The patch essentially removes a bit of code that was covered 100%. Since the overall coverage is below 100%, removing code with 100% code coverage drops the overall coverage since the ration between covered and uncovered code changes in favour of uncovered code.

stanislaw:
I am not sure why you think that the coverage is the same. One of the tests I have written does `     @patch('doorstop.settings.REFORMAT', True)` which makes the following branches to be executed (was not the case before).

```
        # Delay item save if reformatting 
        if settings.REFORMAT: 
            item.auto = False 
```

and 

```
     if settings.REFORMAT: 
            log.debug("reformatting item %s...", item) 
            item.save() 
```

I guess, this is what makes CI to be green now.

sebhub:
Sorry for the confusion, my comment about the coverage was intended for my patch only and not the new tests. Lets say your code base has 10 lines, 5 lines are covered by tests, then your code coverage is 5 / (5 + 5) = 50%. If you remove one line of code which is covered but superfluous, then your coverage is 4 / (4 + 5) = 44%. I think this is what happened with my patch.

stanislaw:
Yes, your explanation makes sense. Thank you.

PR #433, opened 2019-11-01
doorstop/core/builder: .doorstop.skip-all allows skipping subdirs entirely

stanislaw:
This feature is inspired by another pending PR: https://github.com/doorstop-dev/doorstop/pull/431.

While I need this feature to implement LIT-based tests so that they do not collide with the existing test suite, I think this feature can be useful in general because it makes it possible to skip useless directories recursively.

Also I noticed that Doorstop does os.walk through all of the .git, .venv, .cache and other directories which means it does a lot of useless work. I have added a filter for a at least `.git` folder to not be included into the `os.walk` traverse plan.

### DIFF
--- a/doorstop/core/builder.py
+++ b/doorstop/core/builder.py
@@ -36,11 +36,22 @@ def build(cwd=None, root=None, request_next_number=None) -> Tree:
 
     # Find all documents in the working copy
     log.info("looking for documents in {}...".format(root))
-    _document_from_path(root, root, documents)
-    for dirpath, dirnames, _ in os.walk(root):
-        for dirname in dirnames:
-            path = os.path.join(dirpath, dirname)
-            _document_from_path(path, root, documents)
+    skip_file_name = '.doorstop.skip-all'
+    if not os.path.isfile(os.path.join(root, skip_file_name)):
+        _document_from_path(root, root, documents)
+    exclude_dirnames = {'.git'}
+    if not os.path.isfile(os.path.join(root, skip_file_name)):
+        for dirpath, dirnames, _ in os.walk(root, topdown=True):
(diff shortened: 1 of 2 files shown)


doorstop/core/builder.py
```diff
@@ -37,11 +37,24 @@ def build(cwd=None, root=None, request_next_number=None) -> Tree:
     # Find all documents in the working copy
     log.info("looking for documents in {}...".format(root))
     _document_from_path(root, root, documents)
-    for dirpath, dirnames, _ in os.walk(root):
+
+    exclude_dirnames = set(['.git'])
```
jacebrowning: You could also use a set literal here: `{'.git'}`

stanislaw:
Thanks for responding quickly! I need to write the unit tests to catch these coverage bits. Wanted to open this as a RFC first so it is good to have your approval in general!

stanislaw:
@jacebrowning I have written what I think are good tests but the coveralls still says I am decreasing the code coverage.

I am pretty sure it is due to the fact that the number of lines has increased, as @sebhub explained here: https://github.com/doorstop-dev/doorstop/pull/425#issuecomment-542364727.

Any recommendations on how I should proceed?

I could grab any file with the lowest code coverage and open a separate PR to increase the total number. But before I do that I would like to understand where the setting for this check is located: is it a fixed number or the rule is "no less than" that the current value?

jacebrowning:
@stanislaw I just changed the coverage decrease threadshold to 0.1%. I think the tests here are sufficient.

stanislaw:
@jacebrowning it has passed, thanks for the correction. Please proceed with a merge if everything is ok.

PR #438, opened 2019-11-29
Add Python 3.8 to the build matrix

jacebrowning:


doorstop/core/tests/test_all.py
```diff
@@ -610,6 +611,10 @@ def test_lines_markdown_document_without_child_links(self):
             self.assertEqual(expected, text)
         common.write_text(text, path)
 
+    @unittest.skipIf(
+        sys.version_info >= (3, 8),
+        reason="output format differs with newer versions of Python",
+    )
```
jacebrowning: I'm not sure if we're getting a ton of value out of these tests to begin with.

PR #440, opened 2019-11-30
Improve fingerprinting of extended attributes

sebhub:
Do not issue a warning in case a reviewed attribute is missing.  The
doorstop tool has not enough information to do a proper validation of
items of extended attributes.

Make the value serialization independent of the YAML file format.  Use a
simple serialization based on Python string representations.  This helps
in case the item storage format changes in the future.

This is a breaking change, since the fingerprint of existing items may
change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 2.1 (future)
+
+- **BREAKING:** The fingerprint calculation method of reviewed extended
+attributes changed.  Use `doorstop review all` followed by `doorstop clear all`
+to update an existing project. WARNING: This marks all items as reviewed and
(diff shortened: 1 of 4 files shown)


sebhub:
I intensively worked with Doorstop in the last six weeks. I converted the build system of RTEMS board support packages to [waf](https://waf.io) and a build specification maintained by Doorstop:

https://lists.rtems.org/pipermail/devel/2019-November/056267.html

This pull request is a result of this work. The warnings about missing attributes turned out to be impractical.  Unfortunately I missed the 2.0 release by a day and this is a breaking change. Would it be possible to do a 2.1 release if this gets integrated?

PR #441, opened 2019-12-01
Fix fingerprint of reviewed extended attributes

sebhub:
The _convert_to_str() function did not take empty strings, list, and
dictionaries into account.  Use an escape character to encode the type
of a sub-value.

This is again a breaking change which alters the value of existing
fingerprints.

### DIFF
--- a/docs/reference/item.md
+++ b/docs/reference/item.md
@@ -390,4 +390,5 @@ reviewed attributes to the document configuration does not change the
 fingerprint of existing items of the document, if they do not have them,
 otherwise the fingerprint changes.  Removing a reviewed attribute from the
 document configuration changes the fingerprint of all items of the document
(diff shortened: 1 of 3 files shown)


sebhub:
Thanks a lot for the 2.0.post1 release. This is really nice, since it includes all features which I need for the RTEMS project so far. I am really sorry, that I found a new bug in my naive first attempt to serialize values to strings. I would be very happy about a 2.0.post2 release. Sorry for all the trouble.

sebhub:
I hope that it is now good enough.

jacebrowning:
Here's the release: https://pypi.org/project/doorstop/2.0.post2/

sebhub:
Thanks a lot, this is really nice.

PR #449, opened 2020-02-14
Make reference file detection under windows work with linux-created requirements

bfueldner:
Add os.join.normpath in ReferenceFinder to make reference detection of linux-created requirements possible under windows.

### DIFF
--- a/doorstop/core/reference_finder.py
+++ b/doorstop/core/reference_finder.py
@@ -70,7 +70,7 @@ class ReferenceFinder:
         """
 
         log.debug("searching for ref '{}'...".format(ref_path))
-        ref_full_path = os.path.join(root, ref_path)
+        ref_full_path = os.path.normpath(os.path.join(root, ref_path))
 
         for path, filename, relpath in tree.vcs.paths:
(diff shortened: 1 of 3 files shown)


stanislaw:
@bfueldner thank you for the contribution. Would you mind adding a unit test for your case? If that's too difficult, you could simply send me an example of which path you had problems with and I will create a unit test case from it. Just post it here as a comment.

The reason I am asking this is simple: I am not a Windows user so it would be nice to have this case tracked on the unit test level.

bfueldner:
@stanislaw thank you for the review and your suggestions. Writing additional unit tests would not be the problem, but I think all cases are covered in the present tests. The problem is, that the current tests will fail under windows environment, because of some hard-coded path checks and the missing os.path.normpath.

I fix the path asserts in test_item 7fbe9b9 and test_reference_finder 8a48bb4 to make the reference depending tests pass on windows environments.

stanislaw:
Thanks @bfueldner ! As we now have the Window CI enabled by @jacebrowning, I will see if I can fix the rest of the failing tests.

PR #456, opened 2020-02-18
Change python codepage to UTF8

bfueldner:
This patch sets the environment variable PYTHONUTF8 to switch default codepage from latin-1 to UTF8 on windows environments.


jacebrowning:
@bfueldner Thanks for working on this! It looks like there is still one test failure to resolve.

bfueldner:
@jacebrowning Sorry for my long off-time. I close this PR, because it became obsolete with #463.

Issue #452, opened 2020-02-15
Update tests to run on Windows


jacebrowning:
Windows testing was added to the CI suite in https://github.com/doorstop-dev/doorstop/pull/451.

Having these tests pass is not yet a required check to merge PRs, but should be. If anyone has a Windows machine or knowledge of operating system differences please help update the tests so we can make this check required.

stanislaw:
@jacebrowning it is clear from the just merged #449 that the CI now fails on Windows. Looks like any other PRs will fail Windows CI until this issue is resolved. 

I could work on this issue but I cannot promise if I can get to it very soon (probably next weekend).

bfueldner:
@stanislaw I guess the tests fail because of the latin-1 codepage python uses on windows environment as default. As a solution the environment variable `PYTHONUTF8=1` has to be set. I found no `appveyor.yml` in the repo to test it, so the change has to be done manually.

jacebrowning:
@bfueldner the file is here: https://github.com/doorstop-dev/doorstop/blob/develop/.appveyor.yml

jacebrowning:
Closed by https://github.com/doorstop-dev/doorstop/pull/463

PR #462, opened 2020-02-29
Fix failing Windows tests

nhartung:
As @bfueldner states in #452, the 6 tests failing on Windows are related to Windows file encoding. Setting the environment variable PYTHONUTF8=1 fixes the issue and all tests pass. His previous PR was failing CI due to a failing "make check". I committed the auto-format update in order to pass the "make check". Assuming @bfueldner updated the appveyor.yml file correctly, the Windows CI should pass. I also updated CONTRIBUTING.md to add a note that Windows users should set the PYTHONUTF8 environment variable appropriately. 

I have a thought, python allows the command line option "-X utf8" as an alternative to using the PYTHONUTF8 environment variable. This might be preferable if we don't want developers to have to set environment variables to run the tests. I'm not familiar enough with nose to know if it's possible to pass python command line parameters when the tests are run or not. If it is, I'd prefer that approach.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,8 @@
 environment:
   global:
     RANDOM_SEED: 0
+    PYTHONUTF8: 1
+    VIRTUALENV_PYTHONUTF8: 1
   matrix:
     - PYTHON_MAJOR: 3
       PYTHON_MINOR: 6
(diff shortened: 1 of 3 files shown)


nhartung:
Looks like the Windows CI is still failing. I'll take a look to see if the appveyor configuration is being set up properly tomorrow.

nhartung:
Actually, the tests only fail for Python 3.6. The tests for 3.7 and 3.8 pass, but the CI fails on "make demo" for those. PYTHONUTF8 isn't supported in Python 3.6. Maybe an environment variable is the wrong approach to support 3.6.

jacebrowning:
@nhartung I think we can just drop 3.6 from the Windows build matrix for now.

PR #463, opened 2020-03-05
Fixing Windows Tests

nhartung:
The fix proposed in #462 utilized the PYTHONUTF8 environment variable to set the system locale to UTF-8. This approach worked for Python 3.7 and 3.8 on Windows, but PYTHONUTF8 is not supported in Python 3.6. @jacebrowning suggested that we remove 3.6 from the Windows build matrix until we find a solution. 

The above got me thinking. It's clear from the unit tests that the intent is for doorstop to support utf-8, regardless of the system's locale. Given that, I'm not sure that setting PYTHONUTF8 is the appropriate solution. It's also a cumbersome restriction to put on users of the tool. Rather than rely on an environment variable, I decided to track down the source of these failures within the doorstop code. It turns out that a simple update to the read_text and write_text function to specify `encoding=utf-8` in common.py fixes the failing Windows tests.

In addition to the above change, I corrected a separate test whose expected text was relying on unix pathing (Afterwards I realized that this is the same fix that @bfueldner made in #456). I also made some changes to the expected text in tutorial.py (integration tests) to get those passing on Windows as well. It needed to use os.linesep rather than relying on newline characters.

I did a `make test` and `make demo` locally on the following platforms with a passing result:
Windows 10, Python 3.6
Windows 10, Python 3.7
Windows 10, Python 3.8
Ubuntu 18.04 (via WSL), Python 3.8

I'm hopeful that all CI tests are in the green in this PR.

I'm definitely not an expert in the complications of file encodings across platforms using python. I'm open to discussing potential issues with this proposed approach if someone has more experience. If this PR is accepted, we can close #462 and #456.

### DIFF
--- a/doorstop/cli/tests/tutorial.py
+++ b/doorstop/cli/tests/tutorial.py
@@ -238,10 +238,15 @@ class TestSection1(TestBase):
 
         cp = self.doorstop("publish REQ", stdout=subprocess.PIPE)
         self.assertIn(
-            b'1.0     REQ001' + str.encode(os.linesep) + str.encode(os.linesep) + 
-b'        Some text' + str.encode(os.linesep) +
-b'        with more than' + str.encode(os.linesep) +
-b'        one line.' + str.encode(os.linesep),
+            b'1.0     REQ001'
+            + str.encode(os.linesep)
+            + str.encode(os.linesep)
+            + b'        Some text'
+            + str.encode(os.linesep)
+            + b'        with more than'
+            + str.encode(os.linesep)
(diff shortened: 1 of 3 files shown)


bfueldner:
@nhartung Thanks for fixing UTF-8 read/writing at the correct location. I encountered the encoding problem by exchanging doorstop requirements over git between Linux and Windows developers, so I took the *force* switch - which was not the best as I see now!  Aa mentioned above,  I close #456 because the problem is fixed with #463.

Issue #464, opened 2020-03-09
Editing item header in GUI


Entea:
Hey!
First of all, I'd like to thank the project contributors for this awesome software!

I've been poking around doorstop and noticed that there's no way to edit item headers.
Would be nice if it could be done without going to export/import or direct source editing.

Entea:
I've submitted a PR and hope that it can be merged :)

jacebrowning:
Closed by https://github.com/doorstop-dev/doorstop/pull/465

PR #465, opened 2020-03-09
Editing item header in GUI #464

Entea:
Added an input field for item header.


Entea:
Ouch, will fix this soon

Entea:
Hey @jacebrowning! Thanks for taking time and looking into my PR, turns out that I've misplaced the `Links:` row, fixed it and amended into existing commit.

PR #371, opened 2019-07-18
HTML Styling improvements

michaelnt:
These are the styling improvements from rickeywang plus:

- a fix for the missing headings, when an item doesn't have a header attribute
- a fix to the heading levels (but also don't limit to only two level)
- fix the tests (hopefully)

PR #466, opened 2020-03-16
Improve publisher style.

nhartung:
This is #371 with the merge conflicts addressed. @rickeywang does a good job providing an overview of the changes in #353.

I don't have any major comments in addition to what's already specified in #353. I much prefer the format introduced by these changes. The font is better, navigation is cleaner, and the TOC renders more consistently.


doorstop/core/publisher.py
```diff
@@ -386,7 +385,7 @@ def _lines_markdown(obj, **kwargs):
                 yield ""  # break before links
                 items2 = item.parent_items
                 if settings.PUBLISH_CHILD_LINKS:
-                    label = "Parent links:"
+                    label = "🡩 Parents:"
```
rickeywang: @jacebrowning Do you like these arrows?

rickeywang: Is https://github.com/doorstop-dev/doorstop/pull/353#discussion_r300816685 still an issue for you?

doorstop/core/files/doorstop.css
```diff
@@ -1,6 +1,6 @@
 body{
     margin: 0 auto;
-    font-family: Georgia, Palatino, serif;
+    font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
```
rickeywang: See my comments https://github.com/doorstop-dev/doorstop/pull/371#issuecomment-513960855 about this file. I *think* we can get rid of `doorstop.css` or merge it with `general.css`?

doorstop/core/publisher.py
jacebrowning: I don't see an arrow:

nhartung: Strange, feels like a GitHub or browser rendering bug. It renders fine for me using Chrome and Windows 10, but I see the same as @jacebrowning using Chrome on Android. Here's what I see: 
image

rickeywang: Maybe we'll just remove the arrows for now. I put them on as visual candy anyway.

nhartung: It's a question of whether or not there are issues rendering the published documents. I don't think it should be a problem, but as you say, it's nonfunctional. I'll leave it up to @jacebrowning.

doorstop/core/files/doorstop.css
nhartung: I agree. We got rid of the call to _lines_css() which was the only place this css is currently used as far as I can tell. I'll update the PR the remove the file and any unused references to it.

robbel:
@nhartung removing `doorstop.css` results in `FileNotFoundError: [Errno 2] No such file or directory: '/home/philipp/.local/lib/python3.6/site-packages/doorstop/core/files/doorstop.css'` in `doorstop publish` call. There are additional lines to be removed from `publisher.py` (in particular, anything referring to `CSS`) for this to work.

robbel:
@nhartung @rickeywang @jacebrowning I have found the following replacements for the unicode arrows that do render on more machines (e.g., Linux 18.04):
- [Index]: ⇠ ([info](https://www.compart.com/en/unicode/U+21E0))
- [Parents]: ⇡([info](https://www.compart.com/en/unicode/U+21E1))
- [Children]: ⇣([info](https://www.compart.com/en/unicode/U+21E3))
Maybe we can replace the existing arrows in this PR with those.

doorstop/core/publisher.py
```diff
@@ -554,7 +561,10 @@ def _table_of_contents_md(obj, linkify=None):
         elif item.header:
             heading = "{h}".format(h=item.header)
         else:
-            heading = item.uid
+            if settings.ENABLE_HEADERS and item.header:
+                heading = item.header
+            else:
+                heading = item.uid
```
robbel, line 567: This section appears to be superfluous now? `ENABLE_HEADERS` is the default in the mainline version of `publisher.py`, i.e., would propose to exclude this section from the patch.

robbel, line 567: Should these lines be merged into:
```
        elif item.header and settings.ENABLE_HEADERS:
            heading = "{h}".format(h=item.header)
        else:
            heading = item.uid
```
?

neerdoc:
Hi @nhartung , we have recently improved the CI-functionality a lot, so it would be great if you could merge the latest from the develop-branch into your PR.

jacebrowning:
I'm going to close this due to age. Feel free to open a new PR!

PR #468, opened 2020-03-20
Add unique document name check for doorstop create

henrik-garbergs:
The doorstop create command now checks if the specified document name
is unique within the document tree and ends with an error message
if a document with the specified name already exists.

### DIFF
--- a/doorstop/core/tree.py
+++ b/doorstop/core/tree.py
@@ -215,6 +215,14 @@ class Tree(BaseValidatable):  # pylint: disable=R0902
 
         """
         prefix = Prefix(value)
+
+        # Check if a document with the same name already exists in the tree.
+        for d in self.documents:
+            if d.prefix == value:
+                raise DoorstopError(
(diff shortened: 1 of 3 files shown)


doorstop/core/tree.py
```diff
@@ -215,6 +215,14 @@ def create_document(
 
         """
         prefix = Prefix(value)
+
+        # Check if a document with the same name already exists in the tree.
+        for d in self.documents:
+            if d.prefix == value:
+                raise DoorstopError(
+                    "The document name is already in use ({}).".format(d.doc_path)
```
jacebrowning: I don't think we need a new property for this, `document.path` should work.

henrik-garbergs: Fixed. I'm a bit new to git and github, so I might have done a commit too much, see below. But the corrections should be there now. Please tell me how to correct any git mistakes I might have done.

PR #474, opened 2020-04-29
Correction of the clean-install make target 

henrik-garbergs:


jacebrowning:
Does `find $(PACKAGES) -name '__pycache__' | xargs rm -rf` as a one-liner?

henrik-garbergs:
> Does `find $(PACKAGES) -name '__pycache__' | xargs rm -rf` as a one-liner?

Yes, that will work as well. I have pushed that suggestion to this PR now.

PR #476, opened 2020-05-25
Add custom attribute to published output

### LINKED ISSUE #455: Export additional properties
My build system exports requirements to Markdown, then converts them to PDF using panda + eisvogel.

My problem is that for ISO26262 development some requirements need additional properties (e.g.: `ASIL`, `safe_state`, `ftti`) which are very easy to add - I just add a column in the Excel spreadsheet.

However these aren't exported in the Markdown file. As a quick hack I put the `ASIL` (it's only one letter) inside the `summary`, so it is displayed. But the safe state and the fault tolerant time are too long to add in the summary.

I propose to add below the requirement text a list of the user-defined requirement attributes+values.

chrta:
Fixes #455

No tests and no documentation added, since this follows the suggestion in #455.

My questions:
1. Is this approach ok, or should this be implemented in a different way?
2. What kind of tests should be added?
3. Where should the documentation be added?

### DIFF
--- a/doorstop/cli/tests/test_all.py
+++ b/doorstop/cli/tests/test_all.py
@@ -22,8 +22,8 @@ from doorstop.cli.tests import (
 from doorstop.core.builder import _clear_tree
 from doorstop.core.document import Document
 
-REQ_COUNT = 17
-ALL_COUNT = 49
+REQ_COUNT = 18
+ALL_COUNT = 50
 
 
 class TempTestCase(unittest.TestCase):
--- a/doorstop/core/document.py
+++ b/doorstop/core/document.py
@@ -60,6 +60,7 @@ class Document(BaseValidatable, BaseFileObject):  # pylint: disable=R0902
         self.auto = kwargs.get('auto', Document.auto)
         # Set default values
         self._attribute_defaults = None
+        self._attribute_publish = None
         self._data['prefix'] = Document.DEFAULT_PREFIX
         self._data['sep'] = Document.DEFAULT_SEP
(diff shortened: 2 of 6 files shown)


chrta:
@jacebrowning Would this be something that could be merged after adding missing doc and tests? Can you provide some feedback?

jacebrowning:
@chrta Thanks for getting this started! In order see how this feature will work, can you add another `TUT` item with some custom attributes configured to be published? Here's what the [TUT document](https://github.com/doorstop-dev/doorstop/tree/develop/reqs/tutorial) looks like currently: https://doorstop-pr-476.herokuapp.com/TUT.html

jacebrowning:
Thanks for adding that!

Perhaps the attributes table should only be displayed if there are one or more rows in it?

jacebrowning:
Thanks, I like way that published HTML looks! Are you able to rebase on the upstream `develop`? I think that will fix the deployment failure.

> Comment: The tests fail for me (also on my local pc on the develop branch) because of some issue with plantuml.

Would you mind opening a separate issue for this? Maybe someone else can help out!

chrta:
>> Comment: The tests fail for me (also on my local pc on the develop branch) because of some issue with plantuml.

> Would you mind opening a separate issue for this? Maybe someone else can help out!

Sorry, this was an issue when i opened this pr in May, this works now. I will remove this from the initial comment.

Also rebased to current develop.

jacebrowning (approved):
Thanks!

jacebrowning:
This functionality is available in a beta release: https://pypi.org/project/doorstop/2.2b4/

Issue #484, opened 2020-08-04
Utf-8 characters not supported in references[].keyword


Korijn:
**Problem**

Utf-8 characters never match in references[].keyword, I think it's because the pyficache library which is used in https://github.com/doorstop-dev/doorstop/blob/develop/doorstop/core/reference_finder.py#L84 doesn't load files with the utf-8 encoding.

For example, create the following file and save it with utf-8 encoding:

```
français
```

Then try to get the line:

```
>>> import pyficache
>>> pyficache.getline("test.txt", 1)
'franÃ§ais'
```

**Root cause**

When debugging with `pdb`, I can confirm that doorstop uses pyficache which ultimately runs the following code to get the source lines:

```
mode = 'r' if PYTHON3 else 'rU'
with open(path, mode) as fp:
    lines = {'plain' : fp.readlines()}
```

That confirms that this is the root cause (for reasons better explained here https://github.com/doorstop-dev/doorstop/pull/103).

**Proposed solution**

Have you considered using the python builtin `linecache` module instead? It can also do the job just fine and does respect utf-8 encoding:

```
>>> import linecache
>>> linecache.getline("test.txt", 1)
'français'
```

I am aware that pyficache advertises itself as though it also uses the linecache builtin module, but at runtime it (sadly) does not, as my debugging above confirmed.

PR #485, opened 2020-08-04
Item#references: support keywords with UTF8 symbols 

stanislaw:
This is a starting point for fixing #484. 

For now, a test case is added with the `français` keyword as reported in the issue. In the test case, the reference is found without a problem on macOS. Let's see if the test fails on the other systems supported by CI.

Later update: the test indeed was failing on Windows. The solution suggested by @Korijn was to replace the usage of `pyficache` with `linecache`.


stanislaw:
@jacebrowning I need your advice on this. I noticed this while working on the multiple references feature one year ago: all tests work with the shared `files/` folder so that adding any new test fixture (in this case doorstop/core/tests/files/REQ-UTF8.yml) requires modifications in the unrelated tests. This was extremely painful and is still relevant to date.

Would it make sense to create another folder for the test fixtures where they would be maintained "one folder per test case" such as `test_fixtures/001-multiple-references-utf8-keyword` so that no conflicts would be possible anymore?

If I am good to go, I would do this change right away.

stanislaw:
@Korijn, `pyficache -> linecache` doesn't seem to be a trivial change.

```
 File "/.../doorstop/doorstop/core/reference_finder.py", line 48, in find_ref
    lines = linecache.getlines(path)
  File "/Users/Stanislaw/.pyenv/versions/3.6.0/lib/python3.6/linecache.py", line 47, in getlines
    return updatecache(filename, module_globals)
  File "/Users/Stanislaw/.pyenv/versions/3.6.0/lib/python3.6/linecache.py", line 136, in updatecache
    with tokenize.open(fullname) as fp:
  File "/Users/Stanislaw/.pyenv/versions/3.6.0/lib/python3.6/tokenize.py", line 454, in open
    encoding, lines = detect_encoding(buffer.readline)
  File "/Users/Stanislaw/.pyenv/versions/3.6.0/lib/python3.6/tokenize.py", line 431, in detect_encoding
    encoding = find_cookie(first)
  File "/Users/Stanislaw/.pyenv/versions/3.6.0/lib/python3.6/tokenize.py", line 395, in find_cookie
    raise SyntaxError(msg)
Exception: invalid or missing encoding declaration for '/.../doorstop/reqs/assets/logo-black-white.png'
```

I might take a closer look or feel free to check my branch and use it as a base.

jacebrowning:
> Would it make sense to create another folder for the test fixtures where they would be maintained "one folder per test case"?

@stanislaw Yeah, that makes sense to me!

Korijn:
I see that it's failing on a .PNG file, I don't think that needs to be supported, right?

I can have a look and help fix this later.

doorstop/core/tests/test_fixtures/001-item-references-utf8-keyword/REQ-UTF8.yml
```diff
@@ -0,0 +1,15 @@
+active: true
+derived: false
+header: ''
+level: 1.5
+links:
+- REQ001: 35ed54323e3054c33ae5545fffdbbbf5
+normative: true
+ref: ''
+references:
+- keyword: franÃ§ais
```
Korijn, line 10: it looks like you checked in an ASCII file instead of a UTF8 file? it says `franÃ§ais` but I think it makes more sense to test with `français`, right?

Korijn:
By the way, I checked, pyficache indeed returns `None` for a `.png` file.

doorstop/core/tests/test_fixtures/001-item-references-utf8-keyword/REQ-UTF8.yml
stanislaw: You are right! I was also wondering about that but simply copied your example 😄

stanislaw:
The last step is to remove `pyficache` from Poetry's dependency list. I will look into this later this evening because I am having some macOS dev toolchain-related complication warnings as I am running `poetry`.

stanislaw:
> > Would it make sense to create another folder for the test fixtures where they would be maintained "one folder per test case"?
> 
> @stanislaw Yeah, that makes sense to me!

@jacebrowning Thanks! The change is in. This should simplify work with these test cases in the future!

Korijn:
Awesome work! Thanks :)

doorstop/core/tests/test_fixtures/001-item-references-utf8-keyword/REQ-UTF8.yml
```diff
@@ -0,0 +1,15 @@
...
+derived: false
+header: ''
+level: 1.5
+links:
+- REQ001: 35ed54323e3054c33ae5545fffdbbbf5
+normative: true
+ref: ''
+references:
+- keyword: français
+  path: files/text.txt
```
JustinW80, line 11: Should this be `path: files/text-utf8.txt` instead?

stanislaw: @JustinW80, good catch! Technically you are right but the thing is that the existing test does not use this path directly. Instead the full path is passed manually to the `ReferenceFinder` and it is like that by design of this and other similar tests unit tests. Instead of fixing this path, I have pushed a change that explicitly marks the `path: ` as irrelevant for this test. This should help a future reader to see the intention behind this mock-based unit test that tests ReferenceFinder in isolation.

In my opinion, it is very important that this kind of functionality would also be tested with integration tests without any mocking involved as I proposed back then here: https://github.com/doorstop-dev/doorstop/pull/431. Since then I was attracted away to some other projects so I didn't really have time to push for bringing in the end-to-end test suite (be it based on LIT/FileCheck like I suggested there or something manual).

jacebrowning:
I've uploaded this change here: https://pypi.org/project/doorstop/2.2b1/

PR #487, opened 2020-09-24
Add example to examples.md

togaen:
We've begun using Doorstop in the ROS safety working group, and we have a small repo we use for demonstration. We thought it might be a useful example.


jacebrowning (approved):
Thanks!

PR #494, opened 2020-10-09
add machine readable traceability matrix

straun:
In order to enable downstream tooling this creates the same table as in the html index file in a form that can be accessed externally.

When doing a:
```bash
doorstop publish all ./dist/
```

It produces a file called `traceability.csv` by default in addition to the `index.html` file.
This is the _minimal_ feature, it should be possible to add more rich functionality later.


jacebrowning (approved):
Thanks, this will be useful!

PR #496, opened 2020-11-10
Fix #495: missing "self."

### LINKED ISSUE #495: Array references not working on Windows
The "self." is missing from the fourth argument (`self.path`) of the following call in `item.py`
```
reference = self.reference_finder.find_file_reference(
                path, self.root, self.tree, self.path, keyword
            )
```
Additionally, the following approach is suggested for line 73 of reference_finder.py
```
ref_full_path = os.path.abspath(os.path.join(root, ref_path))
```
Using Windows, `path == ref_full_path` doesn't always work unless you use abspath because relative references can result in `\\.\\` being in the path.

This approach was tested using a test file with:
```
references:
- path: tests\\test_filename.py
  type: file
  keyword: test_name_here
- path: tests\\test_filename.py
  type: file
  keyword: test_name2_here
```

knightofni-i:
I see the other issue I brought up in issue #495 has already been addressed, so this PR is very minor.

### DIFF
--- a/doorstop/core/item.py
+++ b/doorstop/core/item.py
@@ -664,7 +664,7 @@ class Item(BaseFileObject):  # pylint: disable=R0902
             keyword = ref_item["keyword"] if "keyword" in ref_item else None
 
             reference = self.reference_finder.find_file_reference(
-                path, self.root, self.tree, path, keyword
+                path, self.root, self.tree, self.path, keyword
             )
             references.append(reference)
         return references


stanislaw:
I will take a look. I would like to find a unit test case for this.

stanislaw:
I would like to merge this change because it definitely fixes the logical error. This error, however, is not something that affects the behavior of Doorstop because the ultimate decision to return a found reference or None is not affected.

[This comment](https://github.com/doorstop-dev/doorstop/issues/495#issuecomment-724943638) questions the efficiency of the current `self.reference_finder.find_file_reference` algorithm so we might also end up rewriting the subject code altogether.

stanislaw (approved):

PR #506, opened 2021-04-18
Validate external references as well as refs

berendkleinhaneveld:
Currently, only the `ref` of an item is validated, while the array of external references is not when running `doorstop. Publishing a document however with non-valid references raises a DoorstopError.

Here is a fix for checking references when running `doorstop`, together with a test.


jacebrowning (approved):
Thanks!

Issue #520, opened 2021-09-02
Text Parsing of Latex Fails on Manual Reorder


wnnmaw:
When performing a manual reorder of requirements, some portion of the requirement text is included in `index.yml` as a comment as a user aid.  However, some text within this comment is incorrectly parsed causing an error.  The following requirements will fail when reordered.

SRD001.yml
```
active: true
derived: false
header: ''
level: 1.0
links: []
normative: true
ref: ''
reviewed: null
text: |
  One requirement
  \begin{equation}
    a^2 + b^2 = c^2
    \sigma = 1.5 \beta
  \end{equation}
```
SRD002.yml
```
active: true
derived: false
header: ''
level: 1.1
links: []
normative: true
ref: ''
reviewed: null
text: |
  Two requirement
```
SRD003.yml
```
active: true
derived: false
header: ''
level: 1.2
links: []
normative: true
ref: ''
reviewed: null
text: |
  Three requirement $Eqn = \frac{a}{b} * \sigma$
```

Console output from attempted reorder

```
(env) C:\Repos\doorstop_issue>doorstop reorder SRD
building tree...
ERROR: editor not found: start
after fixing the error: doorstop reorder SRD

(env) C:\Repos\doorstop_issue>doorstop reorder SRD
building tree...
reorder from 'reqs\srd\index.yml'? [y/n] y
reordering document SRD...
ERROR: invalid contents: C:\Repos\doorstop_issue\reqs\srd\index.yml:
while scanning a double-quoted scalar
  in "<unicode string>", line 16, column 21:
                - text: "Three requirement $Eqn = \frac{ ...
                        ^
found unknown escape character 's'
  in "<unicode string>", line 16, column 62:
     ... quirement $Eqn = \frac{a}{b} * \sigma$"
                                         ^
after fixing the error: doorstop reorder SRD
```

The index file produced by the first call of `doorstop reorder SRD` is 

![image](https://user-images.githubusercontent.com/5656005/131844691-dfde18d5-3540-41c8-a959-c03280cf7cf5.png)

Also note that the line numbering in the traceback is incorrect.

As a workaround, the index file can be modified to delete the offending characters (in this case `\sigma`) from the comment.  The error only appears to occur with specific Latex characters (the `\frac` does not cause an error) and only when an inline formatting is applied.

Versions:
Windows 10.0.19042
VS Code 1.59.1
Python 3.8.11
Doorstop 2.1.4
PyYAML 5.4.1

[...]

PR #521, opened 2021-09-02
Handle backslashes in Reorder operations

asimon-1:
Addresses #520 

Backslashes were not escaped during generation of `index.yml` comments, leading to invalid escape errors when math syntax is used.


jacebrowning (approved):
Thanks!

Issue #524, opened 2021-09-19
Cannot publish item holding special characters in text


urbasus:
Getting an error on publishing a document where an item is having special characters:

- Have special characters in item
  - E.g. paste the MIT license into an item, https://mit-license.org/
  - My name has an "á" in it, not quite the same without it
- Publish

Result:

```
$ make hello/hello.md
doorstop publish -m hello
building tree...
Traceback (most recent call last):
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/common.py", line 115, in read_text
    return f.read()
  File "/usr/lib/python3.7/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa9 in position 196: invalid start byte

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/andreas/.local/bin/doorstop", line 10, in <module>
    sys.exit(main())
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/cli/main.py", line 176, in main
    success = function(args, os.getcwd(), parser.error)
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/cli/commands.py", line 596, in run_publish
    for line in publisher.publish_lines(document, ext, **kwargs):
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/core/publisher.py", line 238, in publish_lines
    yield from gen(obj, **kwargs)
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/core/publisher.py", line 330, in _lines_markdown
    for item in iter_items(obj):
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/core/types.py", line 652, in iter_items
    if is_document(obj):
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/core/types.py", line 628, in is_document
    return hasattr(obj, 'items')
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/core/document.py", line 373, in items
    return sorted(i for i in self._iter() if i.active)
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/core/document.py", line 373, in <genexpr>
    return sorted(i for i in self._iter() if i.active)
  File "/home/andreas/.local/lib/python3.7/site-packages/doorstop/core/base.py", line 178, in wrapped
    self.load()
[...]

PR #526, opened 2021-09-20
Open file using transparent encoding/decoding

urbasus:
See https://docs.python.org/3/library/codecs.html#codecs.open

### DIFF
--- a/doorstop/common.py
+++ b/doorstop/common.py
@@ -8,6 +8,7 @@ import glob
 import logging
 import os
 import shutil
+import codecs
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


urbasus:
Draft state until having added test.

urbasus:
Could use a few pointers on how to test. So far trying to add a file with special characters near other files that I think should be published, starting test with `make ci` which is very slow. No luck. Possibly running the wrong commands, or adding stuff at the wrong places.

- Where to put test?
- How to run test (fast)?

jacebrowning:
https://github.com/doorstop-dev/doorstop/pull/485/files is a good example of adding test files to test encoding issues.

`make test` should run all the tests.

neerdoc:
I've added some tests and sent you a PR on the tests @urbasus.

urbasus:
@neerdoc Amazing work!

Got cold feet regarding this change, suspecting it might be a breaking change. A full environment install later and a whole lot of life events (good and bad) and I never quite got myself to follow through. I'll try get myself to check it out during the week.

neerdoc:
This should solve #543 and #524

jacebrowning (approved):
I think this is fine to merge without additional tests.

PR #544, opened 2022-02-13
overwrite doc defaults by cli defaults

### LINKED ISSUE #530: attributes cannot be overridden by 'doorstop add' command
## Intro
When I set attributes in a .doorstop file, then I cannot use the `doorstop add -d defaults.yml REQ` command to use a different set of default attributes for a newly created requirement.

Basically the `-d` option works as expected, if no default values have been set in the `.doorstop` file.

I'm using doorstop version 2.2b5.

## My setup
.doorstop
```
settings:
  digits: 3
  prefix: REQ
  sep: '-'
attributes:
  !include ../defaults-req.yml
```

defaults-req.yml
```
  defaults:
    header: |
      Please insert an expressive short header here
    text: |
      Please insert your requirement text here
    custom-attribute-1: |
      Please insert a value here
    custom-attribute-2: |
      Please insert a value here
  publish:
    - custom-attribute-1
    - custom-attribute-2
  reviewed:
  - custom-attribute-1
  - custom-attribute-2
```

defaults.yml
```
  defaults:
    header: ''
    text: |
      Please insert your header text here
    normative: false
```

## Run
command line:
```
doorstop add -d defaults.yml REQ
added item: REQ-001 (@\req\REQ-001.yml)
```

## Result
When I look into the item, then I see that instead of applying `defaults.yml`, the `defaults-req.yml` is applied, but the `defaults.yml` has been just copied as a whole in the yml file.

REQ-001.yml
```
active: true
defaults:
  header: ''
  text: |
    Please insert your header text here
  normative: false
derived: false
header: |
  Please insert an expressive short header here
level: 1.0
links: []
normative: true
ref: ''
reviewed: null
text: |
  Please insert your requirement text here
custom-attribute-1: |
  Please insert a value here
custom-attribute-2: |
  Please insert a value here

```

## Expected
A command line option must always override the default values otherwise the option is pointless.

REQ-001.yml
```
active: true
derived: false
header: ''
level: 1.0
links: []
normative: false
ref: ''
reviewed: null
text: |
  Please insert your header text here
```

tangoalx:
fixes #530

### DIFF
--- a/doorstop/core/document.py
+++ b/doorstop/core/document.py
@@ -493,10 +493,14 @@ class Document(BaseValidatable, BaseFileObject):  # pylint: disable=R0902
         more_defaults = self._load_with_include(defaults) if defaults else None
 
         item = Item.new(self.tree, self, self.path, self.root, uid, level=next_level)
-        if self._attribute_defaults:
-            item.set_attributes(self._attribute_defaults)
+
+        # exclusivity: apply defaults given by command line OR
+        # document based defaults, but not both
         if more_defaults:
             item.set_attributes(more_defaults)
+        elif self._attribute_defaults:
+            item.set_attributes(self._attribute_defaults)
+
         if level and reorder:
             self.reorder(keep=item)
         return item


jacebrowning (approved):
Thanks!

Issue #515, opened 2021-06-13
Add templating to publishing functionality


wojciechczerniak:
Hello,

This time a feature request. 

## Problem

I've seen that there are few requests/issues with customization of the Markdown output:

- [x] #166
- [x] #295
- [x] #469
- [ ] #513 
- [x] #314

And I would have few others requirements:

- Start from 2nd lvl header
- Format links as list
- Add custom class to up/down links
- Insert table of contents 
- Various smaller formatting adjustments

But IMO a change from string building to one of the templating languages may solve all of the above in a one go

## Solution idea

I think Jinja2 is the number one for Python libraries? With a template and an ability to use our own fixing all of the above issues would be just a matter of customization with `--template my-markdown.jinja2`

Remove hard-coded string building in HTML publisher:
https://github.com/doorstop-dev/doorstop/blob/a770dc4566a43de87da0e5be61dfedf87a7ac57d/doorstop/core/publisher.py#L165-L167
and Markdown markdown publisher:
https://github.com/doorstop-dev/doorstop/blob/a770dc4566a43de87da0e5be61dfedf87a7ac57d/doorstop/core/publisher.py#L399-L409

With Jinja2 templates defaulting to whatever is in those methods:

```jinja
# {{ document.header }}

{% for item in document.items %}
  ## {{ item.header }} <small>{{ item.id }}</small> { #{{ item.id }} }
  
  {{ item.text }}
  
  {% if item.PUBLISH_CHILD_LINKS %}  
  Links:
    {% for link in links %}
        [{{ link.header }}]({{ link.file }}#{{ link.id }})
    {% endfor %}
  {% endif %}
  
  {% if item.document.publish %}  
        | Attribute | Value |
        | --------- | ----- |
    {% for attr in item.document.publish %}
        | {{ attr }} | {{ item.attribute(attr) }} |
    {% endfor %}
  {% endif %}
{% endfor %}
```

AND allow the `--template` arg for all output formats.

```sh
doorstop publish -m all docs/requirements/ --template my-template.jinja2
```

## Use cases

You want Frontmatter/YAML headers in Markdown #295, just add them to the template:

```jinja
---
title: {{ page.header | title() }}
metadata: {{ page.description | truncate(240) }}
---
{% block content %}
.... content
```

Footer maybe as in #166? Add it at the end:

```jinja
.... content
{% endblock  %}

Copyright Year, Restrictions
```

eterX:
Hi Wojciech

[...]

Issue #295, opened 2018-05-04
Discussion: Markdown with YAML header


dapperfu:
Most Markdown rendering tools allow for a YAML block at the top of the markdown so that you can add metadata to the top of the markdown. Including [Jeckyll and GitHub Pages](https://blog.github.com/2013-09-27-viewing-yaml-metadata-in-your-documents/), [Pandoc](https://pandoc.org/MANUAL.html#extension-yaml_metadata_block), [Hugo](https://gohugo.io/content-management/front-matter/), [Nikola](https://getnikola.com/handbook.html#yaml-metadata), [MkDocs](http://www.mkdocs.org/), among others.

For a proof of concept I took requirement TUT017 and ['converted' it to markdown.](https://gist.githubusercontent.com/jed-frey/cf80624e70742d8333e92183be145d94/raw/f09d3d653eead487f9a48121f2e6c96a01014c26/TUT017.md)

Then used Pandoc with a slightly modified [Eisvogel LaTeX](https://github.com/Voleking/pandoc-templates/blob/master/eisvogel.latex) template to produce this: [TUT017.pdf](https://github.com/jacebrowning/doorstop/files/1971566/TUT017.pdf) with:

```
pandoc TUT017.md \
	--to=latex \
	--output=TUT017.pdf \
	--standalone \
	--template=.pandoc/templates/pdf/Eisvogel/eisvogel.tex \
	--variable=titlepage:true
```

The ```doorstop``` YAML metadata is passed through in Pandoc so custom LaTeX templates can be created for companies to generate PDF requirements documents straight from the ```doorstop``` files.

Second, it would allow using the capabilities existing markdown editors [eg, [remarkable](https://remarkableapp.github.io/), editors with markdown capabilities [eg,  [Notepad++](https://github.com/nea/MarkdownViewerPlusPlus)], or any editor that has markdown syntax highlighting [eg, [vim](https://github.com/tpope/vim-markdown)]. Right now the content is embedded as a YAML field making it easy to mess up since the YAML requires 2 spaces but any Markdown indentation for code needs 4, so code blocks have to have 6 indentations.  With Markdown rendering and syntax highlighting it should reduce mistakes in Markdown formatting.

jacebrowning:
Interesting. The PDF looks great!

Is there an easy way to load Markdown files with embedded YAML **as if** they were YAML files?

It might also be useful to have `md-yml` added as and import-export format.

jed-frey:
[...]

Issue #166, opened 2014-10-15
Add custom Headers/Footers to `publish`


robertberrington:
We have a project that is under Export Restrictions/ITARs ... We need to mark all documentation as restricted.

Currently the `doorstop publish` provides no functionality to mark the "printed version" with any info.

Use case could be:

```
doorstop publish --header="My Company, My Address" --footer="Copyright Year, Restrictions"
```
##

michaelnt:
This is probably addressed by #225 which allows a custom template which could contain the footer.

If exporting to word you could possibly make use of a MD template to add additional information before or after the doorstop data, but probably not a footer. The footer could be added in either the word or latex template of using pandoc

neerdoc:
Closing this issue. This can now be solved with templates for markdown publish or by LaTeX publish.

PR #545, opened 2022-03-02
LaTeX publisher

neerdoc:
After reviewing doorstop for a number of years at different companies I have always gotten stuck on the fact that there is no native publisher for **_LaTeX_**.  Sure, it is _possible_ to go through `pandoc`, but the results have been less that satisfactory.

So, I have finally decided to write a **_LaTeX_**-publisher for doorstop and I hope that it should fulfill everyone's needs.

I have included a simple template for LaTeX so it should be very easy to create nice and customized PDF:s directly from doorstop, so the need for the first part of #295 should be more or less gone. It should also take care of #166 for anyone that is willing to use **_LaTeX_**, as changing the existing template or using a company-template is quite straightforward.

I have also included **_LaTeX_**-packages to allow linking both inside documents and between documents, so that the produced PDFs have a traceability matrix at the end of the document that is clickable even to other documents. Each link to child and parent items are also clickable.

In the same spirit of having a fully customizable template that generates PDFs of high quality with linking between documents, this PR might be useful for the issues #515 and #314.

I have taken care to create a complete test suite for the **_LaTeX_**-functionality, tested on Linux and Mac OS X, and as far as I can see no other tests have broken either.


jacebrowning (approved):
Great work on this! It looks like the only failures are in the output "diff" tests, which I think we can remove because they're not providing a useful failure signal. I'll get a beta release ready so folks can start exploring this feature.

jacebrowning:
This feature is available here: https://pypi.org/project/doorstop/3.0b1/

PR #560, opened 2022-03-08
CI on Pull Request

neerdoc:
Regarding checks not being run on PR I believe is due to the removal of the `on: pull_request: branches` key in the Actions yaml files. Without the `pull_request` key, the checks are only being executed on pushes, which in the case of forks happens on the forked repo and not on the main repo. But with the `pull_request` key installed, the checks should be executed on the fork when a contributor pushes to the fork, and then again on the main repo when the PR is created. It does make an "extra" execution for developers that have the access to push directly to the main repo, but I believe that the benefit of having the tests executed on the main repo trumps that.

This PR is testing this.


jacebrowning (approved):
Thanks!

PR #566, opened 2022-03-12
YAML frontmatter format

ckolumbus:


ckolumbus:
I'm not sure how to deal correctly with the `poetry.lock` file as I'm new to `poetry`` (but love it 👍🏽 ) . Also  this implementation is currently in constant use within my project. I love the preview capability within vscode when editing the `.md` requirement files. the current implementation also synchronizes the `header` field with the first `h1` header in the markdown text. So the title becomes the header.

jacebrowning:
> I'm not sure how to deal correctly with the poetry.lock file

For this project, the easiest way to resolve conflicts is to delete `poetry.lock`, run `make install`, and commit the updated `poetry.lock`.

ckolumbus:
would it make sense to create a real PR from this? Would it have a chance for beeing merged?

doorstop/core/item.py
```diff
@@ -188,7 +189,10 @@ def new(
 
         """
         UID(uid).check()
-        filename = str(uid) + Item.EXTENSIONS[0]
+        fileext = Item.EXTENSIONS[0]
+        if document.itemformat == "yamlfm":
+            fileext = ".md"
+        filename = str(uid) + fileext
```
jacebrowning, line 195: I propose renaming `itemformat` to `extension` (and defaulting to `."yml"`) so this code can be simplified to:

```
filename = str(uid) + document.extension
```

ckolumbus: The whole implementation has changed because the extensions where also used to identify valid files. Just adding to the `EXTENSIONS` array would lead to false positive file identifications (accept `.md` although format was `yaml`).

ckolumbus:
impl, test, doc, rebase done.... one function with quite high complexity, maybe you have ideas on how to do this better. 
The special implementation of heading handling had cost some time. Maybe this could be considered to be refactored.
Check above failed due to connection issues with coverage.io. Can I somehow restart the checks?

ckolumbus:
Is there anything left for me to do before this can be merged? Just let me know!

jacebrowning (approved):
Looks good! I'll get a beta release out for testing.

jacebrowning:
This release contains the new feature: https://pypi.org/project/doorstop/3.0b4/

villanella:
How can I set a document for adding Doorstop items in markdown format with yaml header by default? 
Could I use some  settings like these?
```
// doorstop.yml
settings:
  itemformat: yamlfm
```

Issue #567, opened 2022-03-17
`write_text` has hardcoded line-ending `\n`  - should be OS specific


ckolumbus:
when using doorstop within a development project we have everything is set to `autocrlf=true`. On windows this leads to `crlf` as line ending and on linux only `lf`. `autocrlf` takes care of proper line ending the the working copy.

doorstop is hardcoded to `lf` only in `common.py:write_text` function. This should use the OS dependent `os.linesep` by default to avoid different line ending encodings in repositories.

Additionally: this could be made configurable, but the `end` parameter to the `write_text` function is never used anyway. Maybe this could be adjusted, too. 

Opinions? Feedback?

ckolumbus:
@jacebrowning git has already support for this (stupid) situation, [check this](https://stackoverflow.com/a/40821931) for some more information. when doorstop tries to do it by itself, the git needs to be configured to specifically ignore its own mechanism for this files. 

So my preferred solution would be to have OS specific line endings and have git handle the conversion when working across OSes. 

BTW: cross-OS development happens more often nowadays. I'm a heavy user of DevContainers inside VSCode. So i have my git source code repo (with doorstop requirements) checked out natively on windows and inside DevContainers, which run Linux.

jacebrowning:
> So my preferred solution would be to have OS specific line endings and have git handle the conversion when working across OSes.

👍🏻

ckolumbus:
I can implement the solution, but what would you prefere

1. hardcoded "default" to OS line endings
2. configurable via  document definition file

let me know, then I could make a PR

jacebrowning:
Let go with `os.linesep` wherever `"\n"` is used.

PR #568, opened 2022-04-04
fix: write lines/text with os specific line ending

ckolumbus:
introduced new setting for line endings configuration for `write_lines` and `write_text`. Changes were not done directly in these two functions as this broke a lot of unit-test under windows. If windows as test base is not of importance, then just changing the default parameter for `end` to `os.linesep` might be sufficient.

closes: #567


ckolumbus:
I overlooked the issues last time. Local test on linux run fine, problems might show on windows as a lot of test compare to a golden example ... with `\n` line ending 😦 . 
suggestion for Windows test, in case they fail: configure tests to use `\n` line endings and have one special test added to test the two `write` functions for proper handling of the `end` parameter.

ckolumbus:
tests for correct os  dependent `write_text` and `write_lines`  behavior... then it should be "ready"

jacebrowning (approved):
Thanks for this!

Issue #509, opened 2021-05-17
Doorstep and ISO 26262 / EN 61508


stefanoco:
Also relates to #309 
I'm willing to start a discussion about using doorstop in the context of Functional Safety concerned developments according to mentioned norms. Any experience about this?

jmvillalba:
We are using doorstop in this context. It's usefull to maintain traceability.

But it's necessary to know the limitations of the tool. 

One of the limitations that I miss the most is:
 being able to have several parent documents.
Also, when you change an item, it prompts you to check the traceability upwards, but not downwards.

stefanoco:
Maybe Doorstop will develop in the direction of easy and simple usability, without bothering too much about the peculiar needs of FUSA concerned env, while the similar [StrictDoc](https://strictdoc.readthedocs.io) will evolve more into a FUSA tool? As far as I can see, StrictDoc solves exactly these limitations you mention here

fkromer:
In the past I had to use several tools w.r.t. requirement engineering. One pain point overlooked quite often is interoperability with other tools. Usually requirements of a single compontent need to get into the context of some bigger system. This means the tool you use has to support at least the export of requirements in commonly agreed on requirement interoperability standards which will differ dependent on the domain I guess. @stefanoco Thanks for letting us know about StrictDoc.

ckolumbus:
I do have a proposal for solving the "multi-parent" problem. I'm coming from a medical device software development environment and have similar requirements as mentioned above. We are using Caliber for most of the requirements and the idea is derived from how we define traces there.  It depends on what exactly you need to achieve, but please read on and comment on the proposal

Context:

* Assume we  have the following tree  Feature_Req <- SW_Req <- SW_Test, and now we have a TRA (thread and risk analysis) which "generates" new mitigation requirements based on the Feature_Req.
* I'm also assuming now that the following is sufficient: EVERY requirement from TRA needs to be picked up by the SW_req (implemented)  , but NOT EVERY SW-Req needs to trace to a TRA requirement. In our case this holds true because we need to make sure that all TRA elements are implemented, but not every SW Req actually need one
* on a side note: in order to verify that all TRA requiremetns a tested, a new TRA_Test document could be created as a child to the TRA itself.
[...]

PR #569, opened 2022-04-14
Implementation concept for "mapped" documents to address the multi-parent problem

ckolumbus:
This PR implements the notion of a "mapped" document which addresses (to some extend) the multi-parent problem.

I'm coming from a medical device software development environment and have similar requirements as mentioned in #509 . We are using Caliber for most of the requirements and the idea is derived from how we define traces there. It depends on what exactly you need to achieve, but please read on and comment on the proposal

Context:

* Assume we  have the following tree  Feature_Req <- SW_Req <- SW_Test, and now we have a TRA (thread and risk analysis) which "generates" new mitigation requirements based on the Feature_Req.
* I'm also assuming now that the following is sufficient: EVERY requirement from TRA needs to be picked up by the SW_req (implemented)  , but NOT EVERY SW-Req needs to trace to a TRA requirement. In our case this holds true because we need to make sure that all TRA elements are implemented, but not every SW Req actually need one
* on a side note: in order to verify that all TRA requiremetns a tested, a new TRA_Test document could be created as a child to the TRA itself.
* the main difference to normal tracing in `doorstop` is, that in for normal trace EVERY requirement in a child document needs to trace to a parent req, unless it is derived. Here it is more saying: "no matter what you do elsewhere in the mapped document, you need to pick up all my requirements at least somewhere". This loosens the tight tracing requirements that would  break with *multi-parent* approach. 

If the assumptions are above are valid, then the implementation sketch looks like this

1. let's define a document that holds the TRA: normal document dependencies, child doc of Feature_Req
2. configure an "mapped_to" attribute in the TRA document that lists the documents that need to pick up all the requirements from the TRA, in our case this would list "SW_Req"
3. as a sanity check we could make sure that the mapped document is neither a parent of a child of the current document (direct or indirect, to avoid trace loops) - not done yet
4. when validating document we  check for the existence of the `mapped_to` attribute and validate whether all items from this document are actually picked up by the "mapped_to" document(s). It's not checked, as in the normal case, whether every child document requirement will have a link to this "mapped" parent

Main effort is in the `item_validator`. In addition, a new attribute on document level has been introduced ( and because i need it for something else, also a general "custom attribute" feature similar to items). 

All existing tests are working fine, no tests for the new implementation exist yet. Please review and if this would be a candiate for integration into the upstream repo then the tests and doc updates  follow.

### DIFF
--- a/doorstop/core/document.py
+++ b/doorstop/core/document.py
@@ -186,11 +186,14 @@ class Document(BaseValidatable, BaseFileObject):  # pylint: disable=R0902
                     self._data[key] = value.strip()
                 elif key == "digits":
                     self._data[key] = int(value)  # type: ignore
+                elif key == "mapped_to":
+                    self._data[key] = value.strip()
                 else:
-                    msg = "unexpected document setting '{}' in: {}".format(
-                        key, self.config
+                    log.debug(
+                        "custom document attribute found:  {} = {}".format(key, value)
                     )
-                    raise DoorstopError(msg)
+                    # custom attribute
+                    self._data[key] = value
             except (AttributeError, TypeError, ValueError):
                 msg = "invalid value for '{}' in: {}".format(key, self.config)
                 raise DoorstopError(msg)
@@ -436,6 +439,10 @@ class Document(BaseValidatable, BaseFileObject):  # pylint: disable=R0902
         """Delete the document's index if it exists."""
         log.info("deleting {} index...".format(self))
         common.delete(self.index)
+
+    def attribute(self, attrib):
+        """Get the item's custom attribute."""
+        return self._data.get(attrib)
 
     # actions ################################################################
 
--- a/doorstop/core/item.py
+++ b/doorstop/core/item.py
@@ -728,10 +728,18 @@ class Item(BaseFileObject):  # pylint: disable=R0902
         tree = tree or self.tree
         if not document or not tree:
             return child_items, child_documents
+
+        # get list of mapped documents
+        mapped_document_prefixes = document.attribute("mapped_to") if document else []
+        if not mapped_document_prefixes:
+            mapped_document_prefixes = []
+
         # Find child objects
         log.debug("finding item {}'s child objects...".format(self))
         for document2 in tree:
-            if document2.parent == document.prefix:
+            if (document2.parent == document.prefix) or (
+                document2.prefix in mapped_document_prefixes
+            ):
                 child_documents.append(document2)
                 # Search for child items unless we only need to find one
                 if not child_items or find_all:
--- a/doorstop/core/tests/__init__.py
+++ b/doorstop/core/tests/__init__.py
@@ -4,7 +4,7 @@
 
 import logging
 import os
-from typing import List
+from typing import Dict, List
 from unittest.mock import MagicMock, Mock, patch
 
 from doorstop.core.base import BaseFileObject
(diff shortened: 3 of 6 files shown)


ckolumbus:
@jacebrowning could you please have a look? or let me know who else can help with driving this forward.

jacebrowning:
@ckolumbus I'd recommend pulling in folks from the relevant issues to see if this approach will work for their respective industries.

ckolumbus:
I already asked there,  not much feedback so far .... but I'll try to get some more input on the process aspect👍

From your side any implementation or concept feedback?

stanislaw:
Hi @ckolumbus , here is feedback from someone who is the author of a [similar tool](https://github.com/strictdoc-project/strictdoc). 

The problem that you are solving is very clear to me and it makes complete sense: by making your TRA document's `mapped_to` field to be set to `SW_REQ`, you want to make sure that every TRA requirement is implemented _somewhere_ in the `SW_REQ`, i.e. by the requirement items of the `SW_REQ` ("each requirement of TRA should be implemented by one or more SW_REQ requirements"). 

StrictDoc does not implement this feature yet but I would definitely consider doing it at some point if there was a need.

Here are my thoughts:

- While the `mapped_to` behavior seems clear and practical to me, I don't see how this feature would in any way solve the problem of multiple parents. As far as I understand, the multiple parents problem is related to the fact that you want to be able to link a requirement item to one or more parent requirement items. In my experience, this happens very often that one requirement may refine multiple parent requirements from more than one parent documents (in your example, this would be that a software requirement refines one parent feature requirement and one TRA requirement). As of today, StrictDoc supports child-to-parent `Parent` link type, and I am quite sure that someone should implement multiple parents in Doorstop. ReqIF model also supports `SPEC-RELATION` concept which linking requirements to multiple parent requirements as well as creating custom relation types.

- At work, I don't use Doorstop or StrictDoc (mainly because both tools don't have a powerful GUI which is an absolute must-have for non-software colleagues). Instead, our company uses Confluence for documentation and there we use [YOGI](https://marketplace.atlassian.com/apps/1212523/requirements-yogi?tab=overview&hosting=cloud), a requirements plugin. On top, we put our custom Python automation to export the YOGI requirements database to a large Excel table from which we generate an HTML report where each and every document of our requirements tree has associated statistics where we can see: `Document X: A requirements total, B requirements are refined (C%)` which to some extent ensures that we don't miss any of the X requirements and they will not get refined. This way, we don't have an automatic validation of the `100%` fact like your `mapped_to` would do, but we definitely can see how far are we on the way of refining all requirements that we need.

- If the above comment is valid, i.e. that `mapped_to` and multiple parents are orthogonal to each other, I certainly see both features to solve real-world problems. I don't have time to revisit your implementation of it in Doorstop, but I certainly support concept of `mapped_document` 👍

stefanoco:
Everything good here from my point of view, looking forward at wider adoption in FuSa relevant design processes with this addition.

ckolumbus:
@stanislaw could you maybe provide some more details on the expected behavior of the "multi-parent" mode. 

AFAIU according to the existing doorstop tracing philosophy every child requirement would need to trace to at least one requirement in each of the parent docs. Is this the expected behavior? I'm working in a medical multi-system environment with strict tracing requirements, but I've not seen this specific use case in our process. 

What we rather have is that higher level documents assign specific (==individual) requirements to (sub-)systems. So the 'mapped_to' concept is actually executed on a per-requirement basis, not of full document scop. But again, I believe this is not what you have in mind.

PR #580, opened 2022-07-24
4 publish latex normative section text missing a separating space header and uid

worthenmanufacturing:


worthenmanufacturing:
Another accidental pull request. Sorry.

Issue #497, opened 2020-11-25
Finished documents in lowercase t fail in doorstep-server


jmvillalba:
I have noticed that when I create documents of the type 
doorstop create reqSist ./reqSist --parent ET

And then I run doorstop-server this document is not displayed correctly and shows: 
Sorry, the requested URL 'http://127.0.0.1:7867/documents/pruEs' caused an error:

However the path is correct in the browser
http://127.0.0.1:7867/documents/pruEst

Thank you all for this great tool

jacebrowning:
Fixed by https://github.com/doorstop-dev/doorstop/pull/587

PR #587, opened 2022-12-08
Fixed issue #497

Akoasm666:
The issue stated that doorstop false truncated all files that end with lowercase "t" on its server, it is due to the rstrip(".html") function which deleted all "h, t, m, l"s at the end of a file name. Fixed it by detecting the file ended with ".html" and deleted the particular suffix.

Also, I'd love to open another discussion that doorstop should tell the users their file name cannot be "all", otherwise, the get request will automatically be transferred to the "/documents/all" request (file at doorstop/server/main.py). Another solution would be to delete the "/documents/all" request since it seems to be the same as the "/documents" to me. Moreover, there also should not be any "/" characters existing in the users' file since they will all be truncated as well (although that is not likely to happen).

### DIFF
--- a/doorstop/server/main.py
+++ b/doorstop/server/main.py
@@ -122,7 +122,8 @@ def run(args, cwd, _):
 @hook("before_request")
 def strip_path():
     request.environ["PATH_INFO"] = request.environ["PATH_INFO"].rstrip("/")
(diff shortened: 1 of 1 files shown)


jacebrowning:
This fix is available here: https://pypi.org/project/doorstop/3.0b8/

Issue #327, opened 2019-04-15
HTML Sidebar contents display over top of requirements text when long headings are used


wrc-jasonw:
It appears the sidebar menu that is present in HTML export can display over top of content. In my document the first heading is long. When the page is scrolled all the way up the sidebar wraps text as expected. However when the page is scrolled down the sidebar text extends overtop the page requirements and headings.

Screenshots are shown below.

Screenshot: Text over-top heading when scrolled down:
![image](https://user-images.githubusercontent.com/34869037/56158541-4d667380-5f90-11e9-83d6-c68179000469.png)
Screenshot: Text over-top requirement when scrolled down further:
![image](https://user-images.githubusercontent.com/34869037/56159827-a4217c80-5f93-11e9-9f4a-ee5c94e1d5e1.png)

Screenshot text wrapped when scrolled all the way to top:
![image](https://user-images.githubusercontent.com/34869037/56158509-39227680-5f90-11e9-8525-00f537bff9d6.png)

I propose that the sidebar behavior be modified to accommodate documents with long headings. (to prevent the sidebar text from appearing over-top the requirements/headings)

Browser: Firefox 66
OS: XUbuntu 18.04
Commit: 3c9c4d51d74287eaa8ab0ae6b49deccd170b95ec ("Update dependencies")

wrc-jasonw:
I have done some more investigation.

It turns out the NAV bar gets assigned the "position: fixed;" CSS property. When this happens the element no longer inherits the width from its parent element. Instead it is free to expand  as wide as it wants. I did a google search and found a stack overflow post with a generic solution: https://stackoverflow.com/questions/16018520/is-it-possible-to-keep-the-width-of-the-parent-element-when-position-fixed-is-a

In short: the recommended solution is the use JQuery to dynamically set the width of the <nav id="TOC"> element to the width of its parent. I think should happen when the document is initially loaded and each time the document resizes.

I have a screenshot below showing the element and associated HTML/CSS in Firefox developer mode.
![image](https://user-images.githubusercontent.com/34869037/56206173-bba34880-6019-11e9-924d-73bfd0e64351.png)

rickeywang:
I have fixed this in my own build of doorstop long time ago but forgot to merge it upstream. It's now in #353.

neerdoc:
This is still an issue. Is it possible to fix by adding word-break to the headers?

Issue #589, opened 2022-12-16
Missing assets folder in published output


JustinW80:
Observed on v3.0b8, tested on Windows, macOS, and CentOS.

When I publish to HTML, the output folder contains a templates folder, but no longer contains an assets folder for the style sheets.

The main effect is now viewing HTML published output loses the navigation sidebar and all styling.

## Steps to reproduce:

1. Start with a fresh repo.
2. Create documents and items without using any templates.
3. Enter `doorstop publish -H all output`
4. Open `output` folder.
5. Observe the `assets` folder should exist.
6. View one of the `.html` documents in your browser.

## Expected result: 

An `assets` folder exists in output. The document in the browser should have styled text and a navigation sidebar.

## Observed result

A `templates` folder exists where the `assets` folder should be. The document in the browser is unstyled and missing the navigation sidebar.

## Workaround

Manually rename `templates` to `assets` in the output folder. Then view the documents in the browser.

## Other observations

I think it was first introduced in v3.0b5, possibly from #573 ? I can't quite follow all the templating changes. cc: @neerdoc 

Reverting to v3.0b4, I observed the correct publishing behavior.

jacebrowning:
This beta release contains a fix for this issue: https://pypi.org/project/doorstop/3.0b11/ cc @neerdoc

villanella:
@jacebrowning There is an error in 3.0b12:
```
$ sudo doorstop publish -H all output
building tree...
loading documents...
publishing tree to '/home/username/doorstop/doorstop-rtm/output'...
Traceback (most recent call last):
  File "/usr/bin/doorstop", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.11/site-packages/doorstop/cli/main.py", line 177, in main
    success = function(args, os.getcwd(), parser.error)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/doorstop/cli/commands.py", line 567, in run_publish
    published_path = publisher.publish(
                     ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/doorstop/core/publisher.py", line 85, in publish
    common.write_lines(
  File "/usr/lib/python3.11/site-packages/doorstop/common.py", line 189, in write_lines
    for line in lines:
[...]

PR #612, opened 2023-10-25
Fix html assets

villanella:
#589 and #327 fixes

Details:
For fixing #589 I wanted to copy template dir to output assets dir as a part of get_template function and it worked fine in manual tests, but this change broke some tests, so I decided to duplicate template dir as assets dir in doorstop/core/files instead of code changing.
Maybe the variant from my [first commit](https://github.com/doorstop-dev/doorstop/pull/612/commits/518281c97079d7f6ce4137ac81e7710b6ad7b5b8) would be better, but it requires update of tests.

For fixing #327 (overlapping text by sidebar) I used the idea from this unmerged PR -  https://github.com/doorstop-dev/doorstop/pull/353/files, I added 2 params in sidebar.css:
```
    width: 18em;
    overflow: hidden;
```


jacebrowning:
> Maybe the variant from my [first commit](https://github.com/doorstop-dev/doorstop/pull/612/commits/518281c97079d7f6ce4137ac81e7710b6ad7b5b8) would be better, but it requires update of tests.

I think that would be preferable to duplicating the asset files.

villanella:
> > Maybe the variant from my [first commit](https://github.com/doorstop-dev/doorstop/pull/612/commits/518281c97079d7f6ce4137ac81e7710b6ad7b5b8) would be better, but it requires update of tests.
> 
> I think that would be preferable to duplicating the asset files.

@jacebrowning Thank you for your response, in my MR this is done.

neerdoc:
I think we can close this. #589 is fixed already, and I have a PR soon that fixes #327 as well. I just have to add a few more test cases.

Issue #614, opened 2023-12-19
Latex publisher and custom template not work!


MarcelGep:
Hello together,
maybe i'm stupid but the latex publisher works not with a custom template?

I run
``doorstop publish all ./dist --latex --template PATH/TO/TEMPLATE``

and i get an Error
doorstop.common.DoorstopError: Template flag set, but no 'template' folder was found.

I try a template folder as parameter or a .tex file or a .html file but i get only this error.
What should i set as --template parameter?

(I'm using doorstop v3.0b10)

**Update:**
I look at the source code and the variable "document_template" (template.py) is always "None" because the "template" attribute of a document will never set and the parameter --template is never used only for if/else query.
I don't understand!!?!?!?

neerdoc:
Hi @MarcelGep,

I will look into it. Does it work with other output formats? E.g., if you use --markdown instead of --latex.

Kindly,
Gustav

neerdoc:
Hi @MarcelGep,

the problem is only _very_ bad documentation from my part.

The `--template` flag is ment to set the name of the selected template only. You cannot place it anywhere other than in the `template` folder under the document structure. The reasoning behind that was simply to copy the methodology of the `assets` folder.

So, in your case: `doorstop publish all ./dist --latex --template TEMPLATE_NAME` means that you are actually trying to publish all documents in the standard doorstop folder, i.e., `./reqs`. To correctly supply a custom template, you have to create a folder named `template` under the 'reqs' folder and in that folder place your template files.

```
reqs/
├── REQ001.yml
├── REQ002.yml
├── <snip>
├── assets
│   └── logo-black-white.png
└── template
    ├── TEMPLATE_NAME.cls
    └── TEMPLATE_NAME.yml
```

Please note that you have to supply both a `.cls` file and a `.yml` file and that the `--template` flag takes the name _without_ any file ending.

I hope this helps, and I will try to update the documentation as soon as possible.

Kind regards,
Gustav

MarcelGep:
Ok thanks for the response this way works 👍 
With a correct documentation this will be perfect :-)

MarcelGep:
Ok if i put only the "template" folder into the "reqs" folder it doesn't work! 

> doorstop.common.DoorstopError: Template flag set, but no 'template' folder was found.

It only works if i put the "template" folder into a document folder, e.g. "reqs/sys/template".

jacebrowning:
[...]

Issue #611, opened 2023-10-25
Child/parent links don't work in 3.0b10


villanella:
In Doorstop 3.0b10 all links with anchors has additional .html extension before anchor  that leads to incorrect hyperlink http://mytest.com:7867/documents/parent.html#childlink
The same hyperlink in 3.0b4 is correctly opened.

Workaround in b10: remove *html from the hyperlink or remove `html` extension from `def _format_html_item_link` function in publisher.py:
```
def _format_html_item_link(item, linkify=True):
    """Format an item link in HTML."""
    if linkify and is_item(item):
        if item.header:
            #link = '<a href="{p}.html#{u}">{u} {h}</a>'.format(
            # replace to:
            link = '<a href="{p}#{u}">{u} {h}</a>'.format(
                u=item.uid, h=item.header, p=item.document.prefix
            )
        else:
            #link = '<a href="{p}.html#{u}">{u}</a>'.format(
            # replace to:
            link = '<a href="{p}#{u}">{u}</a>'.format(
                u=item.uid, p=item.document.prefix
            )
        return link
    else:
        return str(item.uid)  # if not `Item`, assume this is an `UnknownItem`
```

But I don't know what is expected format of hyperlinks. The variant without html extension is better for me.
Removing html extension from item.header breaks tests.

jacebrowning:
This beta release contains a fix for this issue: https://pypi.org/project/doorstop/3.0b11/ cc @neerdoc

neerdoc:
Fixed by #615

Issue #601, opened 2023-09-13
Nested lists rendering


villanella:
Doorstop-server can't render nested lists without empty lines.

For example,

    1. Item 1
    2. Item 2
      2.1. Item 2.1
        2.1.1. Item 2.1.1
    3. Item 3

This is rendered as:

    1. Item 1
    2. Item 2 2.1. Item 2.1 2.1.1. Item 2.1.1
    3. Item 3

All nested items are concatenated into one line.

Yet another example:

```
1. Item 1
    1. Item 1.1
    2. Item 1.2
2. Item 2
    1. Item 2.1
```
      
This nested items are rendered as simple, not nested, list:
```
1. Item 1
2. Item 1.1
3. Item 1.2
4. Item 2
5. Item 2.1
```

Expected behavior: support of nested lists made by indents and numbers.
Workaround: to use extra empty lines between items.

Used version - Doorstop v3.0b4 (markdown is default itemformat).

JustinW80:
I think this might be related to #589, where the template theme is missing from versions after v3.0b4.

villanella:
The assets folder exists, the result is the same in my v3.0b4 version and in stable version.

neerdoc:
I'm working on a fix for this now. Please note though, that _correctly_ numbered lists in markdown requires four (4) spaces as indentation, so the first example does not work even for markdown publishing.

jacebrowning:
This beta release contains a fix for this issue: https://pypi.org/project/doorstop/3.0b11/ cc @neerdoc

neerdoc:
Fixed by #615

Issue #581, opened 2022-07-27
LaTeX publish does not respect nested itemize


LarryK82:
Markdown uses indentation to create nested itemizes. The LaTeX Export only creates one
\begin{itemize}...\end{itemize} scope and ignores the indented "*" items

This is caused by the regex searching for "beginning of line followed by * or - or +" and will also affect the creation trigger of \end{itemize} (currently empty line).

Steps to reproduce:
Create a requirement with nested unordered lists and export to HTML and to LaTeX. LaTeX fails (poor Donald E. Knuth will be sad)

`
* asdf
* Abcd
  * xyz
  * QWERTY
`

jacebrowning:
This beta release contains a fix for this issue: https://pypi.org/project/doorstop/3.0b11/ cc @neerdoc

neerdoc:
Fixed by #615

PR #615, opened 2024-01-10
Publisher restruct

### LINKED ISSUE #574: Re-design publisher as own class.
I'm currently looking into all issues on the publishers. I also recently wrote the LaTeX-publisher which now also has templating support. It occurs to me that it would be a much clearer and cleaner code base if publishers were class based.

Should we re-design the publishers to a class-based solution?

neerdoc:
Rather large PR due to changes in the code structure of the publishers to make the code more maintainable, which fixes #574.

While fixing the test cases for the new structure, issues #581, #589, #601, #611 and #614 where also fixed.

Also noted that there are more python versions available, so I added 3.11 and 3.12 and they both pass all the tests so I saw no reason to remove them again.


.github/workflows/execute-tests.yml
```diff
@@ -60,14 +60,22 @@ jobs:
...
-        run: make test
-      - name: Run tests attempt 3
-        if: ${{ failure() }}
-        run: make test
+      - uses: Wandalen/wretry.action@v1.0.20
+        name: Run tests
+        with:
+          command: make test
+          current_path: ${{ inputs.workpath }}
+          attempt_limit: 3
```
jacebrowning, line 68: I suggest rerunning tests at the runner-level rather than restarting the whole suite.

https://pypi.org/project/pytest-rerunfailures/ could be added as a dependency in the future.

jacebrowning (approved):
Thanks for fixing those issues! I'll merge this and publish a beta release for folks to confirm the fixes.

PR #618, opened 2024-01-13
Pre-commit hook

lbiaggi:
Create a doorstop pre-commit hook


neerdoc (approved):
I like it. I already see a use of adding another check for PRs that verify that all changed code has test coverage.

Issue #469, opened 2020-04-02
Markdown labels not compatible with Doxygen


henrik-garbergs:
Hi!
When I publish the REQ document that's included with doorstop, i.e. when I run
`doorstop publish REQ REQ.md`
I get the following result:
```
# 1.0 Overview {#REQ018 }
## 1.1 Introduction <small>REQ019</small> {#REQ019 }
...
```
It would be possible to link to these chapters from other files parsed by Doxygen if it wasn't for the the space between REQ018 and "}". Is this done on purpose in the Doorstop markdown publish or would it be okay if I changed it to work better with Doxygen, i.e. by removing the space?

henrik-garbergs:
The code looks like this, in doorstop/core/publisher.py:
```
def _format_md_attr_list(item, linkify):
    """Create a Markdown attribute list for a heading."""
    return " {{#{u} }}".format(u=item.uid) if linkify else ''
```
If I just change `{{#{u} }}` to `{{#{u}}}` I get rid of the space in the label. I have to update some unit tests for publisher also, but that's easy. Would this be okay to do as a Pull Request?

jacebrowning:
Yeah, I'm not totally sure why that space is there. We'll want to confirm this change doesn't break any of the other export formats.

henrik-garbergs:
The suggested code change is located in a function called `publisher.py::_format_md_attr_list` which is only called from `publisher.py::_lines_markdown`. This function is called when publishing .md-files, which seems fine, and it is also called when HTML files are published. For HTML, the attribute "linkify" isn't set, and because of that `_format_md_attr_list` will return `''` instead of the markdown-formatted string. So from a code walkthrough perspective it looks okay to change this, it will only affect the markdown publication.

I also made a check in practice: I ran the unit tests with and without the code change. The published text and HTML files where not changed:

```
diff doorstop/core/tests/files/published.txt ../doorstop_ref/doorstop/core/tests/files/published.txt
diff doorstop/core/tests/files/published2.txt ../doorstop_ref/doorstop/core/tests/files/published2.txt
diff doorstop/core/tests/files/published.html ../doorstop_ref/doorstop/core/tests/files/published.html
diff doorstop/core/tests/files/published2.html ../doorstop_ref/doorstop/core/tests/files/published2.html
```

The .md files were changed as expected:

```
diff doorstop/core/tests/files/published2.md ../doorstop_ref/doorstop/core/tests/files/published2.md
1c1
[...]

PR #622, opened 2024-01-21
Rework of HTML publisher.

### LINKED ISSUE #26: Add support for Document titles
Such as:

```
document.title = "My Requirements"
```
##

### LINKED ISSUE #327: HTML Sidebar contents display over top of requirements text when long headings are used

### LINKED ISSUE #378: Expected behavior or bug: table of contents of the HTML file is printed in markdown format?
Looking at the `test_lines_html_document_linkify`, it appears that the `publish_lines` method produces the HTML file with a TOC in the markdown format and this looks rather confusing.

This is the file that is checked by this test: `doorstop/core/tests/files/published.html`.

I guess the expected behavior will be to transform this:

```
* [1.2.3 REQ001](#REQ001)
  * [1.4 REQ003](#REQ003)
  * [1.6 REQ004](#REQ004)
  * [2.1 Plantuml](#REQ002)
  * [2.1 REQ2-001](#REQ2-001)
```

into the proper HTML links with anchors.

If this gets confirmed to be a bug, I or anyone else could send a patch.

### LINKED ISSUE #515: Add templating to publishing functionality

neerdoc:
Updated both the HTML-publisher and the `doorstop-server` to user a new layout with a top menu instead of the sidebar, which fixes #327. By upgrading to newer version of bootstrap, the dependency of jquery was removed.

The change also includes the possibility for the users to create really custom HTML exports as the `.tpl` files are now included in the publishing process so that a user can included custom Bottle-template files as well.

Installed a local copy of MathJax so that offline installations work properly.

I also included support for the document attributes that was included for LaTeX publishing, so that the HTML publish also has this information, which fixes #26.

A few bugs that were not yet reported were also fixed, mainly a problem with the #615 PR which had introduced an error when publishing a single document that I did not catch earlier.

Further testing also resulted in reworking the Markdown publishing since links between items where broken, which fixes #378. Since publishing now is separated it was easy to fix without affecting the other formats. This also lead to fixing #469 since the extra space caused havoc to the links.

This PR _might_ also fix #515. After it's been published it would be nice with some feedback on if this is good enough for that issue.

Some dependencies have also been updated and a lot of new test cases to get better coverage on the changed parts of the code.

Also added a couple of new dev-dependencies so that the API server calls can be tested and it is now possible to run `make test-cover` to get immediate feedback on if the changed code is covered by tests of not (instead of waiting for the CI-job to run).


neerdoc:
Will add more tests to cover all changes I made. Don't merge yet... :)

jacebrowning (approved):

PR #642, opened 2024-03-14
Capability to load a different settings.py using a cli parameter

lbiaggi:
As from title to use with different settings with pre-commit & similar cases in CI pipelines

Depends on: https://github.com/doorstop-dev/doorstop/pull/641


lbiaggi:
@neerdoc  can you review this?

PR #617, opened 2024-01-13
Custom item validator per folder

lbiaggi:
Hi @jacebrowning, we have been using doorstop to manage our ISO 26262 requirements, it is part of our core workflow.

We have improved certain areas of the software to match our requirements, and we think it would be useful to share with upstream these changes.

- Load a file as a python module
- A new section inside `.doorstop.yml`  to enable them (we called them extension, as we intend to add more features if necessary) 
       -  Where we can define a custom validator for the items from a specific folder
       -  If file references from this folder should calculate the SHA-256 for them (During review / validate without `-W`)
            -  Change the buffer size for when doing the SHA
- Override doorstop settings from a file through CLI parameter.
 
We made these changes because we didn't want to override most of the default behaviour from doorstop, and yet we wanted to be able to tweak certain validations for specific items.

PR #645, opened 2024-03-21
Generate sha for files in reference

### LINKED ISSUE #564: Feature Discussion: hash based review of external reference content
Hi *, i'd like to get some feedback on a feature idea I have 

Problem:
I like the feature of external references and how (since the new `references` field) multiple ones can be added. Now I wanted to check this references regarding changes. I.e. in case the external content changes a "suspect" or "needs review" validation message comes up so that the external content needs to be reviewed against the requirement again.  With this a tighter connection between the requirement and the implementation can be achieved.

Solution suggestion 1:
* use the "keyword" feature of an ext ref to not only check for existence, but to define a range e.g. via <keyword> .... </keyword>
* have a "stamp" field added to the references items that trackes the defined content
* provide a `reviewref` command to review single references, all references of a req, or all ext references of a doc or doc tree
* make this feature optional so that it can/must be explicitely enabled in the document configuration

Solution suggestion 2:
* keep it independent from the `doorstop` code
* same keyword behavior as above for defining ranges 
* but keep track of the hashes in an extern file or one file for each ext reference (`.doorstop-refhashdb` within the project root, for example)  
* with this no code changes at all are needed

I have implemented both approaches in a POC manner, i.e. no test yet, but basically working. I'll post the link as soon as it's pushed.

lbiaggi:
It introduces a way to let doorstop do the SHA for the files in references.

This depends on: https://github.com/doorstop-dev/doorstop/pull/617 only because it creates the test file.

Closes #564+s.


lbiaggi:
ping @jacebrowning

jacebrowning (approved):
This seems reasonable to me. Thanks!

PR #662, opened 2024-09-30
Fix codecov and test failures

robtaylor:
This *should* fix the issue with linux showing as failing.

### DIFF
--- a/.github/workflows/execute-tests.yml
+++ b/.github/workflows/execute-tests.yml
@@ -59,18 +59,9 @@ jobs:
       - name: Install project dependencies
         run: make install
 
-      # Some tests fails intermittently, likely due to the public runners being
-      # very slow. Especially any client/server tests seems to be problematic.
-      # This is a simple attempt to re-run the tests up to three times if they
-      # fail. Does not add any execution time if successful.
-      - uses: Wandalen/wretry.action@v1.4.4
-        name: Run tests
-        with:
-          command: make test
-          current_path: ${{ inputs.workpath }}
-          attempt_limit: 5
-          attempt_delay: 2000
-          
+      - name: Run tests
+        run: make test
+
       - name: Upload coverage
         uses: codecov/codecov-action@v4
         if: ${{ inputs.os == 'ubuntu-latest' && matrix.python-version == '3.9' && github.repository == 'doorstop-dev/doorstop' }}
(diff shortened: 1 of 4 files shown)


robtaylor:
See https://github.com/doorstop-dev/doorstop/issues/667

robtaylor:
Updated PR to fix two issues causing test and coverage failure:
* codecov required changes to pass the token secret succesfully to the  `execute-tests` workflow
* random changes in plantuml svg generation were seen, so modify publish tests to use a local, known version of plantuml

robtaylor:
Ugh, this changing viewbox parameter in the generated svg is killing me!!

robtaylor:
All working now!

.github/workflows/change-coverage.yml
```diff
@@ -27,6 +27,11 @@ jobs:
           python-version: "3.11"
           architecture: x64
 
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin' # See 'Supported distributions' for available options
+          java-version: '21'
```
jacebrowning, line 33: Why is Java required on CI?

doorstop/core/tests/test_all.py
```diff
@@ -575,6 +586,14 @@ def setUp(self):
         self.tree = core.build(cwd=FILES, root=FILES)
         self.document = self.tree.find_document("REQ")
         self.temp = tempfile.mkdtemp()
+        # Use local plantuml of known version
+        plantuml_cmd = "java -Dplantuml.include.path=includes -jar " + os.path.join(
```
jacebrowning, line 590: I don't think we want to make `java` a dependency of this library. Can you solve this with a Python library?

jacebrowning:
Hi @robtaylor, thanks for getting CI to pass!

Can you open a separate pull request with just the code coverage fixes? As mentioned in the commend above, I don't think we want to add `java` as a dependency in order to get tests to pass. I'd like to review configuration changes and business logic changes separately.

robtaylor:
Hi Jace, yes for sure.

So with respect to plantuml, there isn't really any options - you need to
have java.

That said, as I'm now stripping out the SVG, maybe a better approach here
is to mock calling plantuml. What do you think?

robtaylor:
Hey @jacebrowning, I've put the code changes (plantuml mock) in #669 . This now has just the CI changes.

Issue #667, opened 2024-09-30
using external PlantUML server can break the testsuite


robtaylor:
We don't have any control over changes to the external plantuml server used, so any changes deployed by the plantuml developers can break our test suite, making it fragile. (i.e. http://www.plantuml.com/plantuml see [`core/publishers/html.py:43`(https://github.com/doorstop-dev/doorstop/blob/develop/doorstop/core/publishers/html.py#L43))

Solution is for the test suite to install a version locked plantuml and use this.

PR #669, opened 2024-10-03
Mock calls to plantuml to reduce CI fragility platforms

robtaylor:
As described in #667, calls to and external server for rendering PlantUML introduced fragility to the CI. On the first attempt to fix this issue by including a local copy of PlantUML, it also became clear that it was very hard to get PlantUML to produce exactly duplicate output over different platforms.

This makes the test faster and more robust by simply mocking out the plantuml-markdown execution.


robtaylor:
Fixes #667

robtaylor:
Merged with #670

PR #679, opened 2024-10-16
Add commandline option for publish to exclude the TOC

robtaylor:
This is useful when you want to generate the table of contents externally (e.g. with pandoc), or concatenate the output.


sarnold:
Yes, this would be very useful!  (I've been doing that with `sed`) Actually still trying to mesh the markdown output with Sphinx docs build in the same repo.

PR #719, opened 2025-07-11
migrate from Poetry to UV

danieleades:
migrates from Poetry to UV

UV has several advantages:

- much, much (much) faster
- more actively developed
- standards based, more closely aligned with standard `pyproject.toml` layout.

This PR has several issues, but I wanted to throw it up for discussion before i invested too much time in it-

- I'm not familiar with most of what's in this Makefile. I've possibly broken some things
- there are loads of docs and other cruft that point to poetry that i have not yet updated


jacebrowning:

PR #721, opened 2025-07-12
use 'ruff' for linting

danieleades:
Adds the ruff linter

apart from just giving more lint coverage than is currently configured, ruff is *fast*.

on my machine pylint takes 12 seconds to run on this codebase with a warm cache.

ruff takes 0.32 seconds.

other benefits:

- a big list of optional rules (not yet enabled)
- ruff can also replace both black and isort (formatting not yet enabled)
- ruff has no dependencies (including it doesn't impact dependency resolution)


doorstop/core/publishers/tests/test_publisher_text.py
```diff
@@ -108,15 +107,6 @@ def test_lines_text_item_with_child_links(self):
         # Assert
         self.assertIn("Child links: tst1", text)
 
-    def test_lines_text_item(self):
```
danieleades, line 111: this function was defined twice, with the some content

jacebrowning:

PR #722, opened 2025-07-12
add flake8-comprehensions linting

danieleades:
this PR is stacked on #721


jacebrowning:

PR #734, opened 2025-10-10
make html template contents menu scrollable

cassie-nahar-nasa:
The current HTML template contents menu dropdown is not scrollable if it exceeds the page height. This fixes that.


opt12:
@cassie-nahar-nasa I think this is included in #753.

So it's in `develop` meanwhile.

PR #771, opened 2026-06-03
Issue 770 fix

### LINKED ISSUE #770: Inline Latex not rendering correctly (MathJax Config outdated)
When using inline LaTeX with single Dollars, the math is not correctly rendered with the default template.

```
  ### Math LaTex

  which is rendered like this:

```
renders like 

Image

This is due to an outdated MathJax configuration.

Resolve this issue by updating the MathJax config in the base.tpl file.

opt12:
Fixes https://github.com/doorstop-dev/doorstop/issues/770

- updated the MathJax configuration on `base.tpl` to the new format.
- added an example of inline math in `TUT17.yml 

Now renders correctly:
grafik

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ coverage.xml
 
 # Eclipse
 .settings
+$(git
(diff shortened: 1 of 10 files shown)


opt12:
@jacebrowning Sorry, I fell in the Git trap again.

My base branch was obviously incorrect. I'll start another PR with the correct basis.

PR #772, opened 2026-06-03
updated MAthJax configuration

### LINKED ISSUE #770: Inline Latex not rendering correctly (MathJax Config outdated)

opt12:
Fixes https://github.com/doorstop-dev/doorstop/issues/770

(This is now the correct PR based on the latest develop branch)

- updated the MathJax configuration on base.tpl to the new format.
- added an example of inline math in `TUT17.yml

Now renders correctly:
grafik

### DIFF
--- a/doorstop/core/tests/files/published.html
+++ b/doorstop/core/tests/files/published.html
@@ -6,11 +6,19 @@
   <link rel="stylesheet" href="../template/bootstrap.min.css" />
   <link rel="stylesheet" href="../template/general.css" />
   <link type="text/css" rel="stylesheet" href="../template/doorstop.css" />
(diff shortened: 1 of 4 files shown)


opt12:
I missed to update the test case reference.
I'll add this tomorrow. Please hold on.

opt12:
Test references updated.

jacebrowning (approved):

PR #780, opened 2026-06-25
Fix golden master updates and cleanup handling in tests

### LINKED ISSUE #779: Golden master pattern broken in several tests due to incorrect order and teardown reset
## Description

The golden master pattern is a testing technique where the output of a
function is written to a reference file ("golden master") on the first run.
On subsequent runs the output is compared to this reference file. If the
output changes, the test fails and the reference file is updated, making
the change visible in `git diff` for deliberate review and approval.

This pattern was not working correctly in doorstop due to two separate
issues.

## Issue 1: Assert before write

In some tests the assertion was performed before the reference file was
updated, so the golden master file was never updated on a failing test:

```python
# Broken: reference file is never updated when assertion fails
self.assertEqual(expected, text)
common.write_text(text, path)  # ← never reached on failure
```

The correct order is to write the reference file first, then assert:

```python
# Correct: reference file is always updated, then compared
if text != expected:
    common.log.error(f"Published content changed: {path}")
    common.write_text(text, path)  # ← always executed
self.assertEqual(expected, text)
```

## Issue 2: Teardown reset of golden master files

The test teardown fixture `cleanup_test_yml_files` and `reset_fixture_files`
reset all Git-tracked YAML files after each test, including golden master
reference files such as `exported.yml`. This undid any intentional updates
made by the golden master pattern, so changes to the reference file were
lost before they could be reviewed in `git diff`.

## Expected behavior

- The golden master reference files should always be updated with the
  current output before the assertion is performed.
- Golden master files should not be reset by the teardown fixtures, so
  that intentional changes are visible in `git diff` and can be reviewed
  and committed deliberately.

## Fix

- Correct the order of `write_text` and `assertEqual` in affected tests.
- Introduce a `GOLDEN_MASTER_FILES` set in `doorstop/core/tests/__init__.py`
  and exclude these files from the teardown reset in both
  `cleanup_test_yml_files` and `reset_fixture_files`.

opt12:
closes https://github.com/doorstop-dev/doorstop/issues/779

**Description**

This PR fixes the golden master test workflow.

Changes:
- Ensure golden master files are updated before assertions fail
- Prevent teardown cleanup from resetting golden master files such as `exported.yml`
- Add a heading-type `REQ007` fixture to cover export/publish output changes
- Avoid failing local `test-all` runs when coverage upload is unavailable, e.g. in worktrees

This makes intentional golden master updates visible in `git diff` and keeps local test runs stable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,8 @@ test-int: test-all
 test-all: install
 	TEST_INTEGRATION=true poetry run pytest $(PACKAGES) $(PYTEST_OPTIONS)
 ifndef DISABLE_COVERAGE
-	@ poetry run coveragespace update overall || echo "Coverage upload skipped: no remote origin configured"
+	@ echo
+	poetry run coveragespace update overall
 endif
 
 .PHONY: test-cover
--- a/Makefile
+++ b/Makefile
@@ -111,9 +111,9 @@ test-int: test-all
 
 .PHONY: test-all
 test-all: install
-    TEST_INTEGRATION=true poetry run pytest $(PACKAGES) $(PYTEST_OPTIONS)
+	TEST_INTEGRATION=true poetry run pytest $(PACKAGES) $(PYTEST_OPTIONS)
 ifndef DISABLE_COVERAGE
-    @ poetry run coveragespace update overall || echo "Coverage upload skipped: no remote origin configured"
+	@ poetry run coveragespace update overall || echo "Coverage upload skipped: no remote origin configured"
 endif
 
 .PHONY: test-cover
--- a/Makefile
+++ b/Makefile
@@ -111,10 +111,9 @@ test-int: test-all
 
 .PHONY: test-all
 test-all: install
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -113,8 +113,7 @@ test-int: test-all
 test-all: install
 	TEST_INTEGRATION=true poetry run pytest $(PACKAGES) $(PYTEST_OPTIONS)
 ifndef DISABLE_COVERAGE
-	@ echo
-	poetry run coveragespace update overall
+	@ poetry run coveragespace update overall || echo "Coverage upload skipped: no remote origin configured"
```
jacebrowning: The rest of this diff is good but please undo this change as it's not in scope for the issue. You can already `export DISABLE_COVERAGE=true` to skip this.

jacebrowning (changes requested):

Makefile
opt12: Fair enough!

You are right, this is out of scope and unnecessary.

Maybe you consider to squash the commits when accepting the PR to keep the history cleaner.

Regards, @opt12

jacebrowning (approved):

PR #781, opened 2026-06-26
undo of makefile changes

opt12:
fair enough: the changes in the makefile are really out of scope of the underlying issue.

The Makefile is reverted now.

Mayb3e you squash the commits when accepting the pull request to keep the history cleaner.

Regards!


opt12:
added this PR by accident.It's allready included in #780

PR #782, opened 2026-06-26
Fix heading preservation during manual reorder

### LINKED ISSUE #246: Headings are lost during reorder
For a document ABC that has the following item order:
```
initial: 1.0
outline:
    - ABC001: # Overview
        - ABC002: # Terms and definition
    - ABC003: # This is a requirement
```

And ABC001 and ABC002 are headings (non-normative and level 1.0 and 1.1.0). If I run the reorder command on this and change nothing, ABC002 is no longer a heading. So it seems that doorstop assumes that 1.1.0 is the same as 1.1, which it is in terms of ordering, but it ruins the heading property of the item.

Steps to reproduce:
1. Create the ABC document as shown above
2. Publish the document, notice 1.1 is a heading.
3. Run ```doorstop reorder ABC```
4. Re-publish, notice 1.1 is no longer a heading.

opt12:
**Description**

This PR fixes heading handling during manual reorder and addresses the long-standing issue described in #246.

## What changed

Manual reorder now preserves heading items explicitly in `index.yml`.

Existing heading items are written with a marker:

```yaml
- REQ001: # HEADING # Overview
```

When the index is read back, this marker is recognized and the item keeps its heading level, including the trailing `.0`. This also works for headings without child items, which was the root cause of #246.

The marker is parsed case-insensitively and tolerates additional whitespace, for example:

```yaml
- REQ001: # HEADING # Overview
- REQ001: # heading    # Overview
```

## Relation to #246

Issue #246 reports that heading items without child items lose their heading status during manual reorder because Doorstop previously inferred headings mainly from the index structure.

With this PR, heading information is stored explicitly in the generated `index.yml`, so a heading such as `1.1.0` remains a heading after reorder even if it has no child items.

This should close #246.

## Index comments

When generating `index.yml`, Doorstop now prefers the item `header` as the comment text if one exists. If no header is set, it falls back to the beginning of the item text.

For heading items, the comment is prefixed with `# HEADING #`.

## Documentation

The reorder documentation was updated to explain:

- how heading items are represented in `index.yml`
- how to create new heading items using `# heading #`
- that header text is preferred over item text in the generated index comments

## Compatibility

There are no breaking changes for existing requirement data.

Existing item files do not need to be changed. The new marker is only used in the temporary `index.yml` generated during manual reorder. Existing index files without the marker continue to work as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@
 [![Windows Tests](https://github.com/doorstop-dev/doorstop/actions/workflows/test-windows.yml/badge.svg)](https://github.com/doorstop-dev/doorstop/actions/workflows/test-windows.yml)
 <br>
 [![Coverage Status](https://img.shields.io/codecov/c/gh/doorstop-dev/doorstop)](https://codecov.io/gh/doorstop-dev/doorstop)
-[![Scrutinizer Code Quality](http://img.shields.io/scrutinizer/g/doorstop-dev/doorstop.svg)](https://scrutinizer-ci.com/g/doorstop-dev/doorstop/?branch=develop)
-[![PyPI Version](http://img.shields.io/pypi/v/Doorstop.svg)](https://pypi.org/project/Doorstop)
+[![Scrutinizer Code Quality](https://img.shields.io/scrutinizer/g/doorstop-dev/doorstop.svg)](https://scrutinizer-ci.com/g/doorstop-dev/doorstop/?branch=develop)
+[![PyPI Version](https://img.shields.io/pypi/v/Doorstop.svg)](https://pypi.org/project/Doorstop)
 <br>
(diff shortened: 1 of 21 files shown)


jacebrowning (approved):

opt12:
@jacebrowning Thank you! I very much appreciate, that you still invest a lot of time into this project. 

I try to check whether this also fixes #736.

opt12:
i fixed the merge conflict.
It should work now.
(moved some section in the test by accident which was not automagically recognized.)

jacebrowning (approved):

opt12:
@jacebrowning Just out of curiosity: Do you still plan to merge this pull request?

You approved it, but there was a merge conflict, which is fixed now. 

So there  should be no more obstacle, to merge this PR. Of course, it's up to you.

Cheers!

PR #808, opened 2026-07-26
Do not escape dollar signs delimiting inline math in LaTeX export

### LINKED ISSUE #776: LaTeX Export: Inline HTML not handled correctly as $ gets escaped
After adding some inline HTLM in `reqs\tutorial\TUT017.yml` this works well in the HTML export, but the LaTeX export breaks:

Refer to #770.

```markdown
  ### Math LaTex

  which is rendered like this:

```

is exported to LaTeX like this:
```latex
\subsubsection{Math LaTex}

You can use Math LaTex expressions as \lstinline`$k_{n+1} = n^2 + k_n^2 - k_{n-1}$`
which is rendered like this:\\

$k_{n+1} = n^2 + k_n^2 - k_{n-1}$\\

Alternatively you can also use inline math like this: \lstinline`\$e=mc^2\$` renders as \$e=mc^2\$.
```

As we can see, the Dollar Signs `$` around the inline math gets escaped.

When compiling LaTeX, miktex tries to fix it, but gets it mixed up and hence everything breaks.

## Hotfix

As a first remedy the inline math should be removed from the Tutorial code base.

## Final Fix

Let's have a look on how to prevent the erroneous escaping of single `$` signs in inline math.

Sanjays2402:
Closes #776

Inline math written as `$e=mc^2$` is exported as `\$e=mc^2\$`, which breaks the LaTeX compile. `_latex_convert()` escapes every dollar sign, and the publisher only bypasses it for display math (`$$...$$`).

`_latex_convert_with_inline_math()` splits a line on unescaped dollar signs and converts only the non-math segments, so a complete inline math environment is left alone. Lines with an unpaired `$`, or an already-escaped `\$`, keep the previous escaping behaviour.

The added regression test in `test_publisher_latex.py` fails without the fix and passes with it; the rest of the publisher suite is unchanged.

### DIFF
--- a/doorstop/core/publishers/_latex_functions.py
+++ b/doorstop/core/publishers/_latex_functions.py
@@ -85,8 +85,7 @@ def _latex_convert_with_inline_math(line):
     if len(parts) < 3 or len(parts) % 2 == 0:
         return _latex_convert(line)
     return "$".join(
-        part if index % 2 else _latex_convert(part)
-        for index, part in enumerate(parts)
+        part if index % 2 else _latex_convert(part) for index, part in enumerate(parts)
     )
 
 
--- a/doorstop/core/publishers/_latex_functions.py
+++ b/doorstop/core/publishers/_latex_functions.py
@@ -76,6 +76,20 @@ def _latex_convert(line):
     return line
 
 
+def _latex_convert_with_inline_math(line):
+    """Convert a line to LaTeX, leaving inline math ($...$) untouched."""
+    # Split on unescaped dollar signs so that literal '\$' is left alone.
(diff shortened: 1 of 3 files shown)


jacebrowning (approved):

jacebrowning:
@Sanjays2402 thanks, we just need the code to be formatted.

Sanjays2402:
Formatted in b6cbfae - black wanted the generator expression in `_latex_convert_with_inline_math` on one line. `isort` and `black` are both clean now, and `pytest doorstop/core/publishers/tests/test_publisher_latex.py` is 21 passed.

